PR #5, opened 2015-07-05
Actual data

si404:
CHM_final files set up in the subregion/system/route.wpt format.

For conversion to new format / update to match route changes / etc, while allowing CHM_final to be archived, etc.

Not merging yet, as I don't know how this will affect other sections of the project


jteresco:
I'm unsure if we should commit these old format .wpt files into the new directory structure, or convert to the new file format first.

si404:
Issue with conversion is that we have 12 months of updates in the old format, so would have issues integrating those updates into converted files. Issue with not converting is that we don't have testing on new format yet and that's important to do with data we know works.

Can you convert them without first integrating them with the master branch? What would be the benefit of doing so, rather than converting them in the master branch?

Or we can just put this branch on ice until someone is sure what to do with it.

jteresco:
I think it might be worthwhile to have a period (starting as soon as possible) where we start to bring in fixes to existing systems, still using the old format.  As you point out, this will allow us to bring in things people have worked on over the last many months.  Once we hit a stable point, and before new development, we convert (or decide not to bother).

So I'm leaning toward merging this into the master and opening it up for people to fork and then submit their pull requests for the fixes they've made or will be making.

si404:
Lets start it then! Merging...

PR #11, opened 2015-07-23
Si404 nv update test

si404:
Test update of data using a branch, rather than using my access rights to do it directly (as I've done with eastern Europe).

Added I-11 Future sections and the Summerlin Parkway, as well as minor changes on routes in Arizona and Nevada. Files in these regions now (should) match my local copies.

All done online, as Github for Windows is acting up.


jteresco:
I'll check this out over the weekend.  I see you added a file updates.csv - looks like a good idea.

jteresco:
I need to figure out what the conflicts GitHub reports for this are all about before I can pull.

si404:
It will be updates.csv. The other files haven't been touched. Would it be easier if I re-try this forked approach?

jteresco:
Closing this pull request, as I believe it has been superceded by the newer, larger one.

PR #12, opened 2015-08-05
Update test branch 2

si404:
This one should work Jim.


jteresco:
Does the removal of the presumed US routes through Yellowstone represent a change in policy from what CHM had about how to include routes, or has their official status changed?

Before I bring in the first "big" update, I'd like a few others to have a look.

ovoss:
FWIW, I was in Yellowstone in June, and saw no US route signage within the 
park. But AFAIK that's not a change on the ground, it's long been like that 
but CHM (with some discomfort within the team) went with implied routings 
through the park (US 20, US 89, US 191, US 287) or extending rather 
uselessly into the park without connecting to anything on the other side (US 
14, US 16, US 212).

Other than that I just went through some trouble to clinch those supposed 
routes within the park, the only bother for me is that it would split four 
long US routes. Also, what about US 34 through Rocky Mountain NP (and US 36 
extending into the park to connect to supposed US 34), not included in these 
updates? If we did same there, that would break US 34 in two, with only a 
short stub between US 40 and the western park boundary. OTOH, I don't 
remember whether US 34 and US 36 are signed within the park, or whether 
there are differences from the Yellowstone situation that might warrant 
different treatment.

Oscar



----- Original Message -----

ovoss:
One more concern occurred to me: Removing the US routes (supposedly) within 
Yellowstone will confuse some users who notice their mileage on the affected 
routes is suddenly reduced, or eliminated if their only mileage is within 
Yellowstone. This would remain an issue, even though Si seems to have taken 
care to make sure list files won't break (thanks!), by making in-use 
truncated points alternate labels for the new endpoint (for example, in WY 
US16 the first point is now YelNP +YelstnNP +SylvanPass US20_YNP US89/191 
+US89/191/287 ..., so every eliminated point is moved to the park boundary). 
We don't yet have a good way of communicating to users what is happening, 
and without an equivalent to the Highway Browser yet in place, that makes it 
harder for users to figure out from their log files what's going on.

This suggests to me that if we're going to do the Yellowstone changes, or 
other major updates, maybe we should hold off on those until we have at 
least a partial web "front end", with an Updates page and an HB equivalent, 
to help users identify and deal with those changes. We can still in the 
meantime make minor un-"Updates-worthy" changes to the highway data, such as 
correcting waypoints.

Oscar

----- Original Message ----- 
From: "Oscar Voss" <oscar.voss@comcast.net>
To: "TravelMapping/HighwayData" 
<reply@reply.github.com>; 
"TravelMapping/HighwayData" <HighwayData@noreply.github.com>
Sent: Friday, August 07, 2015 11:31 PM
Subject: Re: [HighwayData] Update test branch 2 (#12)


> FWIW, I was in Yellowstone in June, and saw no US route signage within the 
> park. But AFAIK that's not a change on the ground, it's long been like 
> that but CHM (with some discomfort within the team) went with implied 
> routings through the park (US 20, US 89, US 191, US 287) or extending 
> rather uselessly into the park without connecting to anything on the other 
> side (US 14, US 16, US 212).
>
> Other than that I just went through some trouble to clinch those supposed 
> routes within the park, the only bother for me is that it would split four 
> long US routes. Also, what about US 34 through Rocky Mountain NP (and US 
> 36 extending into the park to connect to supposed US 34), not included in 
> these updates? If we did same there, that would break US 34 in two, with 
> only a short stub between US 40 and the western park boundary. OTOH, I 
> don't remember whether US 34 and US 36 are signed within the park, or 
> whether there are differences from the Yellowstone situation that might 
> warrant different treatment.
>
> Oscar
>
>
>
> ----- Original Message ----- 
> From: Jim Teresco
> To: TravelMapping/HighwayData
> Sent: Friday, August 07, 2015 10:37 PM
> Subject: Re: [HighwayData] Update test branch 2 (#12)
>
>
> Does the removal of the presumed US routes through Yellowstone represent a 
> change in policy from what CHM had about how to include routes, or has 
> their official status changed?
> Before I bring in the first "big" update, I'd like a few others to have a 
> look.
> —
> Reply to this email directly or view it on GitHub.


--

my Hot Springs and Highways pages:  http://home.comcast.net/~oscar.voss/
Hawaii Highways:  http://www.hawaiihighways.com/

si404:
Re Jim: It's not a change in policy - the existence of the routes across the park, when the US route log doesn't have them, nor are they signed, is a violation of the rules. [http://clinched.s2.bizhat.com/viewtopic.php?t=284&highlight=yellowstone&mforum=clinched](Here) is the thread discussing it, with a link to one talking about a 'Select Scenic Highways' system or something as a way of dealing with Yellowstone. Neither discussion really got anywhere.

Re Oscar:
1) good point re: lack of way of communicating a major update. However, given we're still in beta, surely a post to the relevant section of AA Roads would work?
2) Rocky Mountains isn't included as I hadn't heard about that - I thought the only other NP where US Routes didn't traverse, but end at the border, was Great Smokey Mountains.
3) Your clinching work won't be in vain - I have a Yellowstone NP set of routes that has the Great Loop Road, all the access roads and Norris Canyon Road: basically maximum extent of what we had as US routes.

I agree though, that the Yellowstone updates are major enough for discussion before implementation.

Si

ovoss:
Si:

Posting an updates notice to both the relevant regional board on AARoads, 
and a new Updates topic on our temporary board, would help. The latter would 
be a start on an Updates web page down the road once we're out of beta.

One other thing to consider, about drafting new route sets: Let's leave 
something for new team members to do! Still a lot of unclaimed 
state/provincial route sets left in North America (mainly Canadian provinces 
BC, SK, QC, and NL, and many of the southern U.S. states). Much as I'm 
tempted to take on Saskatchewan (traveled a lot there this summer, probably 
even more on another county-snagging expedition next summer), I'll hold off 
on that one for that reason among others.

Oscar

----- Original Message -----

si404:
Oscar - re new systems. This time last year I made the CO and WY state route systems, and Indiana I did last fall. At that point there was no plans for new collaborators. Obviously I'll not make any more, but at the same time, it seems foolish to waste the work I've done.

jteresco:
So anyway, what do we think?  Should I pull in this entire update?  Parts of it?  All except the Yellowstone US routes removal?

si404:
I've undone the Yellowstone changes in the branch as they were the controversial bit, dealt with the recently added issue, and am happy to see the branch go forth like this.

jteresco:
Sounds good.  I should be able to bring it in tonight.

jteresco:
After merge, two minor typos were found, fixed, and committed back to the master on GitHub.

rickmastfan67:
I think we need to revert this (or at least get some fixes done ASAP).  There are several problems with some of the routes.

Case in point, MT US-89.  When did we all agree to change the labeling for mutliplexes of US Highways on Interstates?  I though we were still discussing that part and should have stayed as 'status-quo' till we all decided.
http://www.aaroads.com/forum/index.php?topic=16147.0

All I'm seeing is lines like this with no '+' to denote a hidden label.  Thus, the file is broken completely because 8 lines have double labels without one of them being demoted with the '+'.
337(I-90) I-90(337) http://www.openstreetmap.org/?lat=45.688105&lon=-110.500317

Several other MT US Highways are effected the same way:
US-12
US-87
US-93
US-191
US-212
US-287

yakra:
Alternate labels should be recognized and processed whether they have the '+' prepended or not.

rickmastfan67:
yakra, according to this post ( http://www.aaroads.com/forum/index.php?topic=15733.msg2085696#msg2085696 ), it is indeed broken.

jteresco:
@si404 , do you want me to try to revert MT changes or will you be able to get the issues brought up taken care of?

rickmastfan67:
MT isn't the only state affected it seems.  CO US-40, WY US-191, & WY US-287 are also affected by this and maybe several others that were changed over prematurely before anything was agreed upon.

I just recommend (for now) that we get all the multiplexes of US Highways on Interstates tagged back to the 'current' standard of 'I-79(67)'.  Not '67(I-79)' which hasn't been agreed upon yet.  Or at a minimum, get the '+' back in there since it does break list files for people.

The changes that added exit numbers based off the mileage of US Highways are ok to keep since everybody agreed with that change.

Issue #13, opened 2015-08-08
MA 2 shaping point missing


jteresco:
Reported by deanej on the old CHM forum:

"I noticed that MA 2 has a shaping point that I-91 lacks over the multiplex near Greenfield."

jteresco:
Well, I thought this was closed automatically but I still had to come in to close it.  So I am.

PR #14, opened 2015-08-09
Added shaping point to MA I-91 to match MA 2 along their concurrency.

jteresco:
Writing and testing out the new fork and pull instructions for those who wish to contribute to the highway data.  Also, fixed a broken concurrency, so that's good!


jteresco:
Closes #13 
(Note that the above should also close the issue!)

PR #16, opened 2015-08-09
Update updates.csv

si404:


jteresco:
This looks good.  I'm OK with leaving the "route" entry even though I can get from the DB based on the "root" entry.  I will pull it in later tonight, most likely, when I have time to make the corresponding updates to the site update program and to the updates.php web page.

jteresco:
You know, now that I think of it, we should keep the "route" in there.  In those cases where an updates.csv entry is for a route deletion, it won't have that information there in the DB to be looked up.

PR #26, opened 2015-08-12
ME26 truncation

yakra:
(USA) Maine ME 26: Truncated between the old southern end at Congress Street and the new end at Cumberland Avenue.

### DIFF
--- a/hwy_data/ME/usame/me.me026.wpt
+++ b/hwy_data/ME/usame/me.me026.wpt
@@ -1,5 +1,4 @@
-ConSt http://www.openstreetmap.org/?lat=43.663304&lon=-70.251029
-CumAve http://www.openstreetmap.org/?lat=43.664022&lon=-70.251453
+CumAve +ConSt http://www.openstreetmap.org/?lat=43.664022&lon=-70.251453
 FoxSt http://www.openstreetmap.org/?lat=43.667588&lon=-70.253347
 I-295_S http://www.openstreetmap.org/?lat=43.674805&lon=-70.256310
 I-295_N http://www.openstreetmap.org/?lat=43.679227&lon=-70.256297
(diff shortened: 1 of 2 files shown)


yakra:
Wait, hold up; the Root is missing from updates.csv...

yakra:
OK. Fixed updates.csv. I see that this pull request is now listing 3 commits instead of 2, and I can still see the older state of the file if I follow its link.
Hopefully GitHub is intelligent enough to pull the latest revision of the file...

jteresco:
I'll need to resolve a conflict in updates.csv to merge this, but that's going to happen as multiple people keep adding entries to the top of that file.  I might suggest an alternate where we have a directory of smaller updates files, one per update set, but let's not worry about that right now.  I'll plan to merge this later unless anyone objects.

rickmastfan67:
Or maybe we flip the 'updates.csv' file and add new entries to the end?

yakra:
Either way there's the potential for conflict: Bob's copy has changes that Pete's copy doesn't have; Pete's copy has changes that Bob's copy doesn't have…

jteresco:
The merging will work well in most cases - it's just these files like updates.csv that multiple people will want to be editing in the same position that causes problems.  I'll take care of this one now.

jteresco:
I think this one's set.  Will be running an update soon and we'll see for sure.

ovoss:
Would it be OK, to help with this problem, for me to wait a few days between 
pull requests for new/updated route files and putting in one pull request 
for updates.csv that covers multiple route file changes? I have two issues 
outstanding that will require Updates entries, and more not yet in the 
system, which could result in my rapid-fire submission of multiple route 
file updates. I hope to address my two open issues tomorrow before I go 
offline for the weekend, with others not until next week, but I'd prefer to 
take care of a batch of Updates entries at once next week.

Oscar

----- Original Message -----

jteresco:
I think even better might just be for now to email the updates entries to me to add directly to the master copy when I pull changes in.  There should pretty rarely be conflicts otherwise.  Maybe occasionally on usai.csv or the other multi-region systems.

rickmastfan67:
Or maybe just add the text in the pull request for that file so you can just copy and paste it Jim?

yakra:
> Or maybe just add the text in the pull request for that file so you can just copy and paste it Jim?

yakra:
> Or maybe just add the text in the pull request for that file so you can just copy and paste it Jim?
IAWTP

yakra:
> Or maybe just add the text in the pull request for that file so you can just copy and paste it Jim?

IAWTP

jteresco:
Yes, that would also work.

PR #29, opened 2015-08-12
Minor fix for PA US-322 Business (State College).

rickmastfan67:
Fixed location for 'US322_W' to re-sync route with mainline US-322 (and I-99/US-220).  Also added in one additional point at the most logical road that connects to I-99 Exit #71.  See: http://clinched.s2.bizhat.com/viewtopic.php?t=2388&mforum=clinched

### DIFF
--- a/hwy_data/PA/usausb/pa.us322bussta.wpt
+++ b/hwy_data/PA/usausb/pa.us322bussta.wpt
@@ -1,5 +1,6 @@
-US322_W http://www.openstreetmap.org/?lat=40.816278&lon=-77.929630
+US322_W http://www.openstreetmap.org/?lat=40.815563&lon=-77.928343
 ValVisDr http://www.openstreetmap.org/?lat=40.812250&lon=-77.922506
+VaiBlvd https://www.openstreetmap.org/40.80805/-77.89525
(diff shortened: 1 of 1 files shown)


jteresco:
Looks simple enough, going to pull it in.

jteresco:
Note that the new line used a different OSM URL format than the one our wpt file format supports.  Fixed in master copy, so no worries, but something to keep in mind.

rickmastfan67:
Oops, my bad.  Must have striped out that part by accident when taking out the zoom levels.

rickmastfan67:
It doesn't seem that the fix has been added yet to 'master'.  So you want me to submit the fix myself?

jteresco:
You might need to update your copy of the master to match the one on GitHub in the TravelMapping.  If not, let's worry about it next week.

rickmastfan67:
I double checked the 'master' before I submitted #31.  Last commit was the merge of this ticket with the mistake of mine with the URL string.

I even double checked with what changes were missing between my fork and master.  The URL was still the 'messed' up one for the 'VaiBlvd' point.  So, we'll deal with #31 once your back. ;)

PR #37, opened 2015-08-13
Tch 1 business in chase bc 28

ovoss:
Adding business loop to Chase BC from TCH 1 (Main)

Updates line:

15-08-13;(Canada) British Columbia;TCH 1 Business (Chase BC); bc.tchbuscha;New route


jteresco:
Updates entry made in my local copy for later commit/push.

ovoss:
The Updates entry for BC TCHBusCha is in the updates table, but with an 
extraneous "%20" in the link. The link works once those characters are 
removed.

Part of my latest pull request, with another copy of tchbuscha.wpt (before I 
figured out what was going on), can be ignored.

----- Original Message -----

PR #41, opened 2015-08-18
SC: usasc file location fixes

rickmastfan67:
Corrected the location for a few SC State Highways from the 'usansf' to the 'usasc' folder since they were never part of the 'usansf' system in the first place.  Only SC-22, SC-30, SC-31, & SC-277 are part of the usansf system in South Carolina.


jteresco:
Looks to me like no .csv changes are needed for these.

PR #42, opened 2015-08-18
Md us1 bc tch business

ovoss:
Cleanup on previous pull requests: added missing MD 200 point to MD US1; resubmitting BC TCHBusCha, which is in updates list but not master database.


jteresco:
As you mention later, the BC change was in.   But I think this can safely be merged anyway.

PR #55, opened 2015-08-22
NY point realignments

jteresco:
This is a large update I had sent to Tim long ago but never made it in.  There are a lot of point alignments at intersections, a few hidden points replaced with nearby visible points, and other various cleanup.  I'll pull this in over the weekend unless anyone objects.


ovoss:
Jim:

Unless I missed something in the Southern State Parkway changes, they don't 
truncare the east end of the route to reflect what is now signed as the 
Heckscher Parkway. I don't remember (probably never noticed) where the 
Southern State ends and the Heckscher begins, but most likely at or west of 
the NY 27 interchange.

--

my Hot Springs and Highways pages:  http://home.comcast.net/~oscar.voss/
Hawaii Highways:  http://www.hawaiihighways.com/

jteresco:
Oscar, thanks for the note about that needed update.  This set of changes doesn't address any such updates.  It also puts off I-99 to a later update (which I hope to put in this weekend).  This one is just getting my local files with changes from last summer that never made it to CHM into Travel Mapping.  I'll add an issue here as a reminder for this one.

jteresco:
I think it's safe, so I'm going to pull this in and run an update shortly.

PR #61, opened 2015-08-29
utah activation

si404:


jteresco:
I'd like to see the datacheck for usaut cleaned up (or at least FPs verified to be FPs) before we activate.

PR #62, opened 2015-08-30
Adding I-99, replacing I-99FutMan in NY

jteresco:
Files are a year old on my computer but never made it into CHM.  Pretty straightforward, hopefully.  Updates entries:

15-08-30;(USA) New York;I-99;ny.i099;Added route
15-08-30;(USA) New York;Future I-99 (Mansfield, PA);;Deleted route

I'll pull these in Sunday night or early next week unless anyone notices a problem.


yakra:
For the updates text, instead of saying there was one route added and one route deleted, I'd say that the route was renamed from Foo to Bar. IMO it more clearly indicates that it's the same route, just under a different name.

There are a few different examples on http://cmap.m-plex.com/docs/hwyupdates.php .

Sometimes we have two lines:
(USA) North Carolina US 158 Alternate (Henderson): Renamed to US 158 Business (Henderson).
(USA) North Carolina US 158 Business (Henderson): Renamed from US 158 Alternate (Henderson).
Sometimes just a one-line entry:
(USA) Texas TX Loop 97: Renamed to TX Spur 97.

...Perhaps the single-line "Renamed from" style may be best? That way, the link in the File Root column is a link to the active file in the Region/Route columns.

jteresco:
I agree completely.  Simple but clear update entries are good.  I'm going to pull this in now and run an update.

PR #69, opened 2015-08-30
Alaska state routes (in-dev) updates

ovoss:
Adds AK 98 (once in another in-dev route set), major updates to AK 5 and AK 10 (Chitina), lesser updates to some other state routes.


jteresco:
Will run an update with these very soon.

PR #78, opened 2015-09-02
PA: Adding new interchange for US-322/PA-291.

rickmastfan67:
See: http://clinched.s2.bizhat.com/viewtopic.php?t=2390&mforum=clinched


si404:
Merged

PR #90, opened 2015-09-07
Fixes minor errors in new WI US 41 file

ovoss:
Fixes new US 41 file, to remove alternate labels accidentally carried over from US 45 file.

No Updates entry required.


jteresco:
Will run an update now with this fix.

PR #96, opened 2015-09-09
NYST

si404:
Adds New York State Thruway and Berkshire Connector.

09-09-15;(USA) New York;Berkshire Connector;ny.berconn;New Route
09-09-15;(USA) New York;New York State Thruway;ny.nyst;New Route

I know that this is your wheelhouse, Jim, but taking the 10 minutes to make these files felt worth doing. You are at liberty to change file names, fix them, not include them, whatever - it's your state!


terescoj:
Thanks for these.  I'm very happy to have the Berkshire Connector segment in.  To me it's a no brainer, despite its short length.  The mainline Thruway, I am not sure I'd bother since it's not adding anything we don't already have.

I'm open to opinions, though.

jteresco:
Thanks for these. I'm very happy to have the Berkshire Connector segment in. To me it's a no brainer, despite its short length. The mainline Thruway, I am not sure I'd bother since it's not adding anything we don't already have.

I'm open to opinions, though.

si404:
I guess the mainline Thruway is adding an explicitly signed named freeway to the named freeway system. IMV, it's something that ought to be included, unless there's a good reason not to.

While it adds no new mileage and the file is created, it's merely maintenance, which seems to me to be unlikely to be much work at all. It's almost more bother to not include it, now that I've created the file ;)

ovoss:
I also would not bother adding a file for the mainline NY Thruway as a named 
route. That doesn't add anything to our covered mileage, and adds hassles 
whenever we need to update route files for the concurrent Interstate routes.

On the Berkshire Connector filename, BerConn -> BerCon, or better still -> 
BerkCon? If we're going to add a strictly-unnecessary letter, "Berk" seems 
more useful than "Conn".

Oscar

----- Original Message -----

si404:
BerCon sounds better - I had assumed, wrongly, that other connectors had the four letter variant.

USAIF doesn't add any mapped mileage as a system, AFAICS - every bit is a numbered US, Interstate, or state highway, or a named KY Parkway. But it creates an extra layer of data useful for the end user.

yakra:
While the AR I-49Fut segments are state highway, they're not included in USANSF -- USAIF only.

I vote for BerCon. (I think I may have a "Conn" label or two still hanging around ME US1 from the very early days that I haven't gotten around to changing. Bad Yak!)

I know that, out there in the field, there's a bit of the mainline thruway (reference route 915H) that's neither I-87 nor I-90. But whether this is actually represented with the lo-res nature of our maps I'm uncertain. Away from PC; can't check.

si404:
WRT to AR I-49Fut (and SC I-73/I-74Fut) they could be covered by USANSF, and will be covered by USAAR (and USASC), and that they are the only ones you use - there might be some others (CA905, for instance), but there's really not a lot that aren't duplicates that add no extra mileage. Of course, 

My point is that having these future interstate routes (many unsigned) says that it has been agreed that they add something, despite being concurrent with other mappable routes. So why not explicitly-signed-as-such Named Freeways like the New York State Thruway?

Obviously it's not my place to insist that this route is entered - I commented the entry for the NYST out when I changed the name of the Berkshire Connector in the .csv files, knowing that (while there) it wouldn't be added - but I do feel that the excuses as to why not are rather weak.

jteresco:
I'd like to bring in the Berkshire Connector segment, but I haven't decided if I should make it just the short segment between the mainline Thruway and B1, or extend all the way to the MA line.

yakra:
I just looked at the NJTP for guidance. Well sonofagun -- even in the CHM HB, it includes the entire NJTP, not just the "(South)" bits it started out as.

How'bout extending all the way to the MA line, then?

ovoss:
I agree. Especially since the other Bx exits on I-90 suggest that part of 
I-90 came from the original Berkshire Connector.

At the other end of the connector, I suggest the west endpoint be labeled 
I-87, rather than NYST.

----- Original Message -----

jteresco:
OK, going to add the whole segment.  Rather than pulling in this one partially, I'm just going to make changes in my own copy and close this once done.

jteresco:
The Berkshire Connector is now in the master repository and in the TravelMapping DB.  Closing this issue.  If we want to have further discussions about whether to add the remaining Thruway routes (which I am leaning more in favor of doing) we can discuss that in the forum or in a new Issue here.

PR #107, opened 2015-09-20
Alexander ND bypass

yakra:
Can't automatically merge.
I think because: I made a commit to updates.csv on the 15th, then waited for my previous pull request to be merged to avoid wacky hijinks. Then Si made another commit to updates.csv (on the 16th?), which was merged. My new commit today incorporates the changes he made, so things should be OK...
But let's look this one over, shall we?

15-09-20;(USA) North Dakota;US 85;nd.us085;Removed from US 85 Business and relocated onto the western Alexander bypass.
15-09-20;(USA) North Dakota;US 85 Business (Alexander);nd.us085busale;Added route.
15-09-20;(USA) North Dakota;ND 200;nd.nd200.wpt;Removed from US 85 Business and relocated onto the western Alexander bypass.


jteresco:
Don't worry about the merge troubles - I can resolve the conflicts I'm sure.  Going to try that right now, since this seems like a pretty straightforward update (other than the updates.csv conflict).

PR #112, opened 2015-09-28
Wahoo bypass

yakra:
15-09-28;(USA) Nebraska;US 77;ne.us077;Removed from 1st St, Chestnut St, 12th St, and CR 16 and relocated onto the northwestern Wahoo bypass, between NE 92 West and NE 109 South.
15-09-28;(USA) Nebraska;NE 92;ne.ne092;Removed from 1st St, Chestnut St, 12th St, and CR 16 and relocated onto the northwestern Wahoo bypass, between US 77 South and NE 109.


jteresco:
Site update complete with this change.

PR #115, opened 2015-09-30
PA: Some minor PA fixes

rickmastfan67:
http://clinched.s2.bizhat.com/viewtopic.php?t=2256&mforum=clinched (re-sync PA65TrkMcK with PA51)
http://clinched.s2.bizhat.com/viewtopic.php?t=2373&mforum=clinched (added in missing connection for PA-351 to I-76/I-376 interchange)

Changes don't require any mentions for the update page.


jteresco:
Just launched a site update that should have these in the system shortly.

PR #120, opened 2015-10-03
Minor Tweaks

si404:


si404:
15-10-03;(UK) Wales;A465;wls.a465;Route removed from road through Rassau and onto new expressway (both called Heads of the Valleys Road) between A467 and A4048
15-10-03;(UK) Wales;A4046;wls.a4046;Route extended at north end from A4047 to A465
15-10-03;(UK) Wales;A4281;wls.a4281;Route extended at north end from Old route of A465 to A465 (new route)

jteresco:
This one's in.

PR #132, opened 2015-10-10
LA US371

yakra:
15-10-10;(USA) Louisiana;US 371;la.us371;Removed from LA 177 and US84 and relocated onto a new two-lane east-west roadway between the LA 177 north split and the US 84 / LA 1 junction.

-----

I lined WPTedit up to the NAIP imagery and did a little Alt+Tabbing to aid in waypoint placement.


jteresco:
Looks straightforward enough - merging.

PR #127, opened 2015-10-06
Watford City ND bypass

yakra:
15-10-06;(USA) North Dakota;US 85;nd.us085;Removed from US 85 Business and relocated onto the southwestern Watford City bypass.
15-10-06;(USA) North Dakota;US 85 Business (Watford City);nd.us085buswat;Added route.
15-10-06;(USA) North Dakota;ND 23;nd.nd023;Removed from ND 23 Business between US 85 Business and the southeastern Watford City bypass, and relocated onto the southeastern Watford City bypass between US 85 and ND 23 Business.
15-10-06;(USA) North Dakota;ND 23 Business (Watford City);nd.nd023buswat;Added route.
15-10-06;(USA) North Dakota;ND 200;nd.nd200;Removed from US 85 Business and relocated onto the southwestern Watford City bypass.

PR #133, opened 2015-10-12
Watford City ND bypass

yakra:
Old US85 points added as discussed in #127


jteresco:
No additional updates or modifications for the updates.csv file from this?

yakra:
Correct. All of that was taken care of with the previous pull request.

jteresco:
OK, great.  Launching an update now.

PR #136, opened 2015-10-16
Update datacheckfps.csv

si404:
Added in Utah FPs, and changed some European ones that were lost due to either label changes (x~3), additional shaping points (x1) or (which never used to happen) moving a point slightly (x lots!)


jteresco:
Going to run an update now with these, looking forward to a shrinking list of FPs!

PR #140, opened 2015-10-16
Si404 patch 2

si404:


jteresco:
Nice!  That sure shrinks our datacheck list.

PR #152, opened 2015-10-20
TX datacheck

yakra:


jteresco:
This batch is live in the DB.

PR #153, opened 2015-10-21
TX In-Dev system datacheck FPs 

yakra:


jteresco:
Texas is now clean of non-FP datachecks!

PR #158, opened 2015-10-26
MS/TN: I-269

rickmastfan67:
Here's the brand newly opened I-269 files for both MS and TN.  Thankfully, MS doesn't have a State Highway system yet, so, didn't have to mess around with those files since the MS section ends @ a state highway and could add it safely.  Just wanted to have the boarder points to be synced.

Now, I'm not sure if they have added I-269 shields to TN-385 yet, but, let's just work under the assumption that only the brand new segment has them for now.  Thus why I extended the Future file of I-269 down to the new interchange instead of merging it into the new I-269 file.

Info for the update file:
15-10-25;(USA) Mississippi;I-269;ms.i269;Added route.
15-10-25;(USA) Tennessee;I-269;tn.i269;Added route.
15-10-25;(USA) Tennessee;I-269 Future (Memphis, TN);tn.i269futmem;Extended southward from the interchange with TN 57 to a newly opened interchange with I-269 (Exit 2).


jteresco:
Update running now with these included.

PR #168, opened 2015-10-28
AK98 removed from in-dev CANSPH set

yakra:
superseded by AK98's inclusion in USAAK


yakra:
With this done, I believe we can uncomment cansph from systems.csv, and add it to the DB as in-dev.
Shall we give it a go?

jteresco:
Already done in this afternoon's update!

PR #180, opened 2015-10-31
NL & TX in-dev edits

yakra:
no update texts, as these systems are in-dev.


ovoss:
A global question, unrelated to NL2: Should those of us maintaining other 
provinces be updating their Canada Select Provincial Highways and Canada 
Select Named Freeways routes? My inclination, for BC, SK, and QC, would be 
to leave their updates (I know of some in SK, at least) for whoever ends up 
taking on those unassigned provincial route sets, which might not be me.

--

my Hot Springs and Highways pages (NEW ADDRESS):
http://www.alaskaroads.com/home.html

si404:
Oscar, I would say that, while you are under no obligation to, it would be nice to do what you can there, so that if you do hand it over, someone else gets routes that are up-to-date.

yakra:
Agreed. It'd be nice, but needn't be mandatory or even compulsory.

yakra:
Derp. I meant, "or even a priority."

ovoss:
"It'd be nice" isn't enough for me. There are two in-dev systems on my plate 
already (a limit I'd stick to, especially since one of them is the large 
California state route set), not to mention other things both TM and 
otherwise, to keep me busy and out of trouble for awhile.

Oscar

----- Original Message -----

yakra:
With the system all but officially mothballed, I also see no problem with letting the routes sit & gather dust till someone's ready to take them on as part of a full provincial set.

Issue #188, opened 2015-11-03
Wisconsin BUS US 53 Duplicacy


the-spui-ninja:
So the files for US 53 Bus in both Superior and Solon Springs are exactly the same.

This seems like this isn't right.

ovoss:
I'll submit a pull request in a few minutes, to replace the US 53 Business 
(Solon Springs) route file with one copied from the CHM database.

----- Original Message ----- 
From: the-spui-ninja
To: TravelMapping/HighwayData
Sent: Monday, November 02, 2015 10:31 PM
Subject: [HighwayData] Wisconsin BUS US 53 Duplicacy (#188)


So the files for US 53 Bus in both Superior and Solon Springs are exactly 
the same.
This seems like this isn't right.
—
Reply to this email directly or view it on GitHub.



--
Oscar Voss - oscar.voss@comcast.net - Arlington VA

my Hot Springs and Highways pages (NEW ADDRESS):
http://www.alaskaroads.com/home.html

jteresco:
Update running now with this fix.  No harm done on the extra file in #189.  Backed out with #192 and committed only the needed file directly to master.  Closing.

PR #189, opened 2015-11-03
Responds to issue #188

ovoss:
Replaces erroneous route file for WI US 53 Bus (Solon Springs) with copy from CHM database.


jteresco:
Looks like this mistakenly brought in a change to usaca_con.csv also.  I'm backing out the merge and will copy in the Solon Springs file separately.

PR #211, opened 2015-11-18
Hi 380 relocation retry

ovoss:
Short part of HI 380 closed and relocated to tie route into new Airport Access Rd. (route signage unknown). Waypoint for access road added to HI 36 route file.

Update entry (needed only for HI 380):

2015-11-17;(USA) Hawaii;HI 380;hi.hi380;Route relocated from closed Dairy Rd. segment between waypoints AirAccRd and PakSt_N, onto parts of Airport Access Rd. and Pakaula St.


ovoss:
Ignore this pull request, I'll submit a new one with correct new files.

PR #242, opened 2015-12-04
VA updates

jteresco:
VA updates received by email from Mapmikey.  Please review changes to active systems before I pull these into the master and use in a site update.


mapmikey:
VA228truck is weird.  Coming from either direction on 228 the truck route is posted to use the western half.  But the eastern half is fully posted (and the eastern half is where 228 either direction pointed you to in the past, per GMSV) in both directions.  If the protocol is to then include the eastern half I will put it together, though someone should let me know how to do a loop without crashing the system.

ovoss:
To do a loop, add waypoints for the eastern half of the loop from VA228_N 
except for the one where the loop rejoins VA228_S. Then copy the first line 
with coordinates for VA228_S, change the copy's waypoint name to something 
else (I suggest EldSt), then add the renamed line to the end of the file. 
That will probably generate a "duplicate coordinates" Datacheck error, which 
should be flagged as a false positive and crossed off the Datacheck error 
list. That's similar to what was done for GA I-285 and NC I-485.

Oscar

----- Original Message -----

PR #253, opened 2015-12-11
NM NM369 fix

ovoss:
Fixes Datacheck sharp angle errors, from misplaced CR 7 waypoint. Part of in-dev system, no update entry needed.


yakra:
Weirdsville, daddy-o...
When I look at the diff, the before and after versions of the file are identical, save for a terminal line feed. The before-version is the only other version in TM/master, also identical. No out-of order points.
But if I look in datacheck or in the HB, the sharp angle errors are right there, plain as day.
What is going on...

jteresco:
OK, I'll hold off on this.  I'll also re-run last night's updates to make sure the version on TM/master is the same as the version used in the update.

ovoss:
The NM 369 Datacheck items are new ones, so what's on the master isn't the 
old version of the file.

Looking again at my pull request #251, it looks like the NM 369 waypoint 
order somehow got scrambled. The version in pull request #253 has the 
waypoint order right (US84/285_N, CR7, NM581, US84/285_S), so using that 
should fix the problem.

----- Original Message -----

jteresco:
Still not showing up right in the HB.  For some reason the master and my copy don't match the commit here.  I'm making the change (swapping first 2 lines of my version of the file) and will commit back.

ovoss:
Jim:

That worked, Datacheck errors for NM 369 cleared.

There are Datacheck errors for other NM state routes, which should all be 
false positives. But first I'll double-check all the other recent file 
changes, to make sure none of them went awry.

Oscar

----- Original Message -----

yakra:
Strange. I see the error in #251 as Oscar describes above.
But still, how did it make it into the DB/HB without ever getting into master?

All I can think of is that XKCD comic Jim posted in another thread...

jteresco:
Good in the HB now.  Definitely an odd one.  Hopefully the error won't creep back in somehow.  Datacheck will make sure we notice if it does.

PR #255, opened 2015-12-12
Final NM SR changes

ovoss:
Last few route file updates needed before usanm activation.


yakra:
NM118: should AltLabels even be in use if the system hasn't been activated yet?

ovoss:
I've used AltLabels sparingly for the pre-activation usanm route set. In this case, to avoid list file breakage from point label renames just a day or two before activation. 

----- Original Message -----

jteresco:
I don't mind the AltLabels being in there.  They add a near-zero cost to site update processing, and no cost at all to the time or space needed after processing.

I'll pull this in and run an update.  If things look clean, I'll activate.

jteresco:
Looking at the forum, I'll wait for word about the possible partial decommissioning mentioned there before activation.  Update running now.

PR #256, opened 2015-12-13
Really final NM SR changes

ovoss:
Final (really) round of changes to in-dev New Mexico state routes before activation, to reflect recently decommissioned routes/segments, as discussed in the forum.


jteresco:
Running an update with these now.  If all looks good, I'll re-run the update with usanm as an active system unless anyone objects.

jteresco:
Was removal of NM 456 intentional?  I didn't notice anything on the forum about that one going away.

jteresco:
OK got it - I'll put it back.

PR #258, opened 2015-12-16
Usanyp

si404:
Splits off New York Parkways into own system.

Hopefully I've got them all


jteresco:
If you'd like, I can move the files from the NY/usasf directory into NY/usanyp at the command line when I pull this in.

Any objections to this coming right in as an active system, even though we'll be developing some new routes for it?  Any preference for keeping the existing routes where they are, developing new routes in usanyp, then once the system is complete, move the old usasf routes into usanyp?

ovoss:
ISTM the best reason to keep usnyp as an in-dev system for now is that we 
can add rough drafts of new parkway route files to the system, and make as 
many changes as needed without a bunch of Updates entries until we're 
satisfied they're ready for "prime time". Duke87's comment on the Saw Mill 
Parkway route file, which quickly went live but now looks like it will have 
to be immediately revised (along with the Henry Hudson Parkway file), 
suggests to me that further additions to the parkways system would benefit 
from at least a brief stay in "in-dev" status.

Otherwise, if already-activated files are available and fully mappable as 
part of one system or another, I don't see how it matters to users which 
system they are in.


----- Original Message -----

yakra:
> "Any preference for keeping the existing routes where they are, developing new routes in usanyp, then once the system is complete, move the old usasf routes into usanyp?"

Glad you mentioned this. Yes, I like this approach. It allows for some peer review before activating a bunch of new routes. Just from adding two, we received some comments on both.

jteresco:
Given the comments so far, I'm going to hold off pulling this one in.

PR #262, opened 2015-12-16
CT I-395: sequential -> milepost

yakra:
No .list files are affected.
No updates entry necessary.


rickmastfan67:
I think when a highway has a complete renumbering, a mention should be made for the update page.  We've done that before when we've had to 're-enter' a highway because of the entire highway getting new numbers.

Something simple like this would have worked IMO:
Converted highway exit numbers to brand new mileage based ones from previous sequential based exit numbers.

Do you agree with me on this @jteresco?

jteresco:
No harm putting in an update entry.  Those with the old numbers will see an entry in their log saying that the old point was replaced, but might as well inform those who might just want to fix up their numbers right away.

yakra:
Agreed, no harm in an update entry.

jteresco:
I am putting in an update entry that will show up later tonight on the site.

PR #264, opened 2015-12-18
USACA removals, USANP additions

ovoss:
Removes some mileage within national parks from in-dev USACA system. That splits two routes, and truncates others.

Most of that mileage, and other roads, in seven routes added to in-dev USANP system.


ovoss:
The change to a file in chm_final was inadvertent, Please ignore that part of the pull request.

Since this is the second time this week I've let such an accidental change into a pull request, is there a way for me to modify a pull request myself after submission?

jteresco:
I think that as the author of a pull request you can make whatever changes to it you wish.  When I've made these kind of mistakes, I've just closed the bad pull request and started another. But I think you might be able to go into the web interface to the pull request, under commits or files changed, and remove part of the request.  But I say if it takes more than 5 minutes to figure out that route, just create a new pull request without that chm_final change.

ovoss:
I've tried to figure it out, this is proving to be a huge headache (even on 
undoing the bad commit without losing all the wanted changes I made). I can 
try again this evening.

----- Original Message -----

yakra:
Perhaps just a new commit to revert the bad one to its previous version?

ovoss:
I did, and I seem to have restored the files I had included in my previous 
pull request to the versions I had tried to pull (which I then copied to a 
non-GitHub directory, for safekeeping). Only when I create a new branch, it 
says "no local changes", so no obvious way to get the restored files into 
the new branch for a new pull request, to replace the old request I couldn't 
edit. GitHub blocked some of my earlier attempts to change branches, saying 
that they would've erased other work in progress.

I poked around in the online GitHub Help, didn't find an item addressing how 
users can edit their pull requests to remove unwanted or erroneous 
components.

Jim, would it be possible to restore the pull request I closed, with you 
undoing the change to chm_final I had wanted to edit out? Or I could e-mail 
files to you for direct addition to the GitHub master. Then after that's all 
done, I could delete all the branches I had created in an attempt to get 
this to work, and synch my copy of the repository to the GitHub master, so I 
can start with a clean slate for my next round of file changes.

GitHub is a very convenient way to make changes when done right, but seems 
to get user-hostile on correcting mistakes.

Oscar

----- Original Message -----

jteresco:
I'll pull this in and restore the (one?) file that was mistakenly changed.  I think your idea of matching up your copy to the master before the next round is a good one.

jteresco:
Should be good now.  I replaced the chm_final version of usaca.csv with a subsequent commit.

PR #267, opened 2015-12-19
Fixes part 2 for usaca

ovoss:
This just fixes usaca csv (no change needed for usaca_con.csv), to fill in missing Abbrev field for two recently-added routes.

This does not affect any files in previous Fixes part 1 pull request (#266), and both can be processed together. It does affect a csv file merged into the GitHub master a few hours ago, which I reviewed and spotted the errors this pull request is to fix.

While I think there should be no conflict, a GitHub message below says there is one, which I can't figure out.


jteresco:
The changes here are simple enough that I'll just make them in the master instead of trying to resolve the conflict.  Should be done in a few minutes.

PR #273, opened 2015-12-21
Region csv files

si404:
.csv files for the region system. May not be 100% accurate (eg continents for grey areas, geopolitical issues), and is certainly subject to change.

### DIFF
--- a/continents.csv
+++ b/continents.csv
@@ -1,0 +1,8 @@
+code;name
+AFR;Africa
+ASI;Asia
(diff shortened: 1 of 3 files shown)


jteresco:
Thanks - looks like CO and DC snuck into the country list.  I'll pull this and and remove those.

jteresco:
These are in, with DC, CO, and a couple others removed.  I should be able to get them into a few new DB tables today.

PR #275, opened 2015-12-21
CA I- and US routes partial update

ovoss:
Part of general cleanup of I- and US routes, especially in far northern and eastern CA. Rest of cleanup still in progress.

No Updates entries needed. Most notable change is addition of points to I-80 route file from MQOpenSat imagery, where new Bay Bridge alignment peels away from old alignment, for anyone (not me) who wants to count the bridge replacement as a de-clinch.


jteresco:
Somehow ca.i080.wpt and ca.i505.wpt were moved from usai to usaib.  I'm putting them back, assuming that was unintentional.

PR #279, opened 2015-12-24
French N roads

si404:


si404:
Adding in some other tweaks based on construction openings:
2015-12-23;Austria;E55;aut.e55;Removed from B310 and relocated onto S10 between S10/B310 junction and B125
2015-12-23;Italy;E35;ita.e35;Removed from A1 and relocated onto A1Var between northern and southern junctions between A1 and A1Var

PR #290, opened 2015-12-30
CA 4 and related updates (in-dev)

ovoss:
Reroutes CA 4 in Stockton and Angels Camp; adds new CA 4 Business in Angels Camp; and makes conforming changes to CA 49 and CA 99 route files.

PR #292, opened 2015-12-31
Ca 4 updates redux

ovoss:
Fixed filename for ca.ca004busang, that foiled pull request #290


jteresco:
I didn't notice it was just the missing "00" in the filename, or I'd have just renamed it.  Thanks for the quick resubmit.  Looks like all's OK now.  Update running.

PR #295, opened 2016-01-02
SC: US-521 Connector removal in Sumter, SC.

rickmastfan67:
This route should have been deleted at the same time when I rerouted US-521 in Sumter, SC along the 'connector's' former route.  This pull request fixes that problem.

2016-01-01;(USA) South Carolina;US 521 Connector (Sumter);;Route deleted


rickmastfan67:
Just realized I forgot about the 'US-15 Connector' that followed the same route and needs to be axed too. :facepunch: Submitted for removal in #327.

PR #305, opened 2016-01-03
Old forum issues

yakra:
Sorry, no time to write the required updates text entries now. I'll add them in a comment later tonight.


jteresco:
All in devel systems, right?  So no updates entries necessary unless I'm missing something.  I'll pull these in.

yakra:
Whoops. HA! You're right: no updates entries necessary.

PR #320, opened 2016-01-06
Old forum issues part 3

yakra:
2016-01-06;(USA) Kentucky;US 60;ky.us060;Removed from the old demolished Ledbetter Bridge and KY 6074, and relocated onto the new Ledbetter Bridge about 1/2 mile southeast, between US 62 in Reidland and KY 6074.
2016-01-06;(USA) Kentucky;KY 16;ky.ky0016;Removed from KY 3716 and relocated onto Pride Pkwy between the southern and northern KY 3716 intersections.


yakra:
Paducah fixed in #320

PR #324, opened 2016-01-07
usanes activation, and KY873Hec

yakra:
Activates usanes, and adds KY873 (Hector).
Updates text is committed directly to the updates.csv file.


jteresco:
Trying an update with this now.  Thanks.

PR #329, opened 2016-01-08
Update oh.oh072.wpt

mapcat:
Changed waypoint name for former OH794


jteresco:
Looks good, thanks.

mapcat:
It worked! Cool! I just sent another one (last one for now). In the future,
is there a limit to how many files you would want in one request?

Also, what kinds of changes need to be reported in the Highway Data Updates
page?

Thanks!

jteresco:
People tend to group together from a few up to a couple hundred file changes in a single pull request.  I think it's best to group related changes together into a single request, but it would be nice to have completely unrelated updates in separate pulls.

The only things that need to be in the highway data updates are changes to routes in active systems that might cause someone's .list file to break or otherwise need an update.

PR #343, opened 2016-01-10
Super fun system explosion!

si404:


jteresco:
Picky detail: Which is right?  MAR in countries.csv or MOR in systems.csv?

si404:
MAR is correct - I've just changed it all locally. It's the Morocco/Maroc issue and the standardisation people's love of French (and Morocco is rather Francophonic)

PR #345, opened 2016-01-10
160110

mapcat:
Hope this did it...


si404:
Looks like you are getting the hang of this github malarky, It's not easy!

yakra:
Looks good; many commits listed together in one pull request. I'll look over the changes a little later when I'm back home at my PC.

mapcat:
Yeah, I'll see if I remember the steps next time I have a lot of updates!

I already know of a change I need to make to one of the files. Is there a way to commit a new version of the file to this pull request, or should I just wait until this one is merged and create a new one?

yakra:
If you make a commit to the same branch this pull request is from, it will automatically be added to this pull request.

mapcat:
Ah, right. I thought it would have replaced the first oh002 but it added the updated version at the end. That's not going to cause a conflict, is it?

yakra:
A-OK, No conflict.

jteresco:
Essentially you have two commits that include the same file, they are applied to the original file in the order you committed them, so all should be fine.

I'll be pulling this and some other items in later today.  Seems to be quite a bit of activity.

yakra:
To add on to my earlier comment:
Now this this pull request is merged, future commits to the mapcat:160110 branch won't be included in *this* pull anymore. You can keep on adding commits though, and just make a new pull request for those.

PR #346, opened 2016-01-10
MOR -> MAR

si404:


si404:
I didn't rename mor.a008, see https://github.com/TravelMapping/HighwayData/pull/347

PR #352, opened 2016-01-13
Ohio updates 2016-01-12

mapcat:
noteworthy:
RonReaHwy: new route (usasf)
I-275: waypoint 31->31A
I-275: new waypoint 31

rest:
adding exit numbers to US6, US24, US422, OH10, OH44, OH129, OH176
correcting waypoint positions/names

Sorry for the bogus commits ae35c60, 710f1d0, 823a946, and 821543c but I do not know how to get rid of them (they do not affect any files afaict)


jteresco:
Thanks - could you make updates entries in the format in the csv file?

mapcat:
Done. Thanks.

yakra:
Looking thru US24, I noticed OH424C & AntWayTrl_W were hidden in place, with a +.
Better off marking them as closed intersections instead, with an asterisk.
(I can't type in an example, because GitHub flavored markdown will convert to either a bullet point or italic text.)
There may be similar examples in other files I haven't looked at yet.

yakra:
Also noting that although a RonReaHwy file was created, it's not added to the usasf CSV files yet. Was this intentional?

jteresco:
No - I just added it.  The file was also misnamed.  Commit and push coming momentarily.

mapcat:
@Eric: the +OH424C was already in the file in that format, so I guessed that +AntWayTrl_E was the way to go with that closed intersection. Actually since AntWayTrl_E wasn't in use, I wanted to just make it +(number) but didn't know if there was a way to automatically renumber the other hidden points. Is there? Is there a reason to preserve names for closed intersections other than to offer them as points for those who join in the future who may have used them before they were closed?

@Jim: Sorry about forgetting the oh. on the RonReaHwy file...was that the only issue?

yakra:
> Is there a reason to preserve names for closed intersections other than to offer them as points for those who join in the future who may have used them before they were closed?

Well, that's the *best* reason. And I can't think of any others, much. Other than the historical context of "there used to be an intersection here; it's closed now." And that's just, well, useful for someone coming along in the future.

It's possible someone could make heavy use of closed intersection points as the route changes out from under their travels. A good example from my .list file is [NB1](http://tm.teresco.org/devel/hb.php?r=nb.nb001&u=yakra) -- 8/16 points used are marked as closed, and 7/8 lines contain a closed intersection. At 47.6%, there's not much left of my original clinch c. 1989.

jteresco:
Re: @Jim: Sorry about forgetting the oh. on the RonReaHwy file...was that the only issue with the name?

That's the only issue I saw with the name.

PR #354, opened 2016-01-16
KY Parkways and other changes

si404:
2016-01-16;(USA) Kentucky;Hal Rogers Parkway;ky.halrogpkwy;New Route

Adds requested HRP, splits ky parkways off usasf, gets roudn and myse to devel level, and does some minor tweaks for nlda0


yakra:
LOL, I had just drafted my own HalRogPkwy in WPTEdit, but not yet made a commit.
I'd done a couple things differently: KY638 point included, and some minor intersections to use visible instead of hidden shaping points, and break up long visible distances. Any objections if I make a PR later on to add in these edits?

Also: I merged this PR in then remembered about the idea of using USANYP as a holding ground for Preview parkways, then activating them as USASF once they've had some peer review.

Shall we revert this PR, include just HalRogPkwy in as preview USAKYP system, and then once satisfied it's good to go, move all the parkways over & activate USAKYP?

si404:
My draft of HalRogPkwy has been there for a while. Certainly improve my file if you feel that there's issues - I have no objections.

I did toy with the idea of having usakyp as a holding place like with usanyp. I would not be opposed to that commit being reverted and that different set up set up. Do as you see fit (just remember to unhide roudn in systems.csv).

yakra:
At this point, the holding place setup is probably not worth the bother of reverting this PR, changing around all the CSVs, and then changing around the CSVs again once HalRogPkwy gets reviewed. Let's keep your commits and go forward from here.

The only potential snag I see (a la HenHudPkwy/SawMillPkwy) is, where's the western terminus?
Wikipedia seems to indicate it's at US25. Signage at that junction only has it going east. But there's also signage at I-75...

yakra:
OSM calls the unlimited, developed bits of KY80 between I-75 & US25 Russell Dyche Memorial Hwy. Historic Aerials shows it predating the Parkway there by quite a bit. I'm going to call the end US25/80.

mapcat:
Actually now the HRP is even longer, ending at Somerset (via KY 80):

Stretch of Ky. 80 renamed for Hal Rogers
http://www.sentinel-echo.com/news/stretch-of-ky-renamed-for-hal-rogers/article_f97aa4cc-7694-11e5-8947-87d2abc0f06e.html

yakra:
Crap, uhh, okay, I'm gonna read up on that...

yakra:
extended in #356

jteresco:
I'm out some tonight, so I can't do it immediately, but if the dust has settled on this I'll pull them down (along with other recent items) and run an update tonight.

PR #356, opened 2016-01-16
KY tweaks

yakra:
Following up on #354.
~~DO NOT MERGE THIS IN YET.
I still have some more commits to add.~~
Safe to merge now.


yakra:
Safe to merge now.

Since the site hasn't been updated since #354 was merged, I see no point in having updates text for the truncation-then-extension of the HRP.

Issue #332, opened 2016-01-08
KY: Clinton County notes


yakra:
reposition ky.us127busalb south end
KY1076 south end, minor reroute

yakra:
US127 & US127BusAlb fixed in #364.

Leaving this issue open, as US127 still needs a little TLC in southern KY.
• points on old alignment, e.g. KY558

mapcat:
@yakra sorry for missing this. I made some slight realignments to US127 to match up with newer imagery. Anything else necessary?

yakra:
KY558 is still off-center; that's about it really

mapcat:
Ah, I hadn't checked that far north. Fixed in my files.

mapcat:
Live in HB.

yakra:
Shall I close the issue?

mapcat:
If you consider it solved, then yes, please.

PR #362, opened 2016-01-19
KY1076: relocation at south end, to TN state line

yakra:
system in preview status; no updates text necessary

addresses part of #332


yakra:
b9be010 follows up on #361.

PR #366, opened 2016-01-21
Mostly devel system changes

si404:
Other than TUR D22Duk (whose update is already there), all this is in devel


jteresco:
I should be able to pull these in tonight and run an site update.

PR #368, opened 2016-01-22
adding in-dev cannf

yakra:
Nota Bene: I have made the following changes to roots in @ovoss's regions:
sk.skcir.wpt -> sk.cirdr.wpt (Circle Drive)
bc.gew.wpt -> bc.golearsway.wpt (Golden Ears Way)
I haven't changed ON RR174. Discuss.


rickmastfan67:
Well, with Ontario RR174, this would be the best place for it since all the other 'numbered' Canadian routes will be merged into each respective provencial system.  I don't think it makes sense to have a system for a single route.

That file for RR-174 is only for the 'Queensway' part.  If it hadn't been for all those Ontario downloads, that might still be part of ON-17.

yakra:
Should the route & root be Que? Or keep or as RR174?

Issue #30, opened 2015-08-13
CA Interstates cleanup


ovoss:
Reviewing all CA Interstates for needed changes such as for new interchanges, or waypoints out of synch with other Interstate route files. Updated locally so far: I-80 (includes waypoints for where new Bay Bridge peels away from old roadway, for people unlike me who consider the new bridge a de-clinch) , I-505.

rschen7754:
The Ranchero Road interchange on I-15 (exit 140) is now open. Same for Duncan Canyon Road (exit 117) and Nisqualli Road (exit 148).

ovoss:
Thanks for the heads-up. I'll make the changes to the I-15 route file, and 
include that in a pull request covering all the routes with new 
interchanges, etc. (I-80 and I-505 updated so far).

Oscar

----- Original Message -----

PR #371, opened 2016-01-23
CA Interstates cleanup part 2

ovoss:
Addresses part of issue #30 (cleanup of the I-5 and I-10 families, and I-x80 routes, to come later). CA US 95 is affected by I-40 tweaks, and its revised route file is also included. Updated in-dev CA CA299, on which I fixed a weird Datacheck error, is coming along for the ride.

Update item for only route with new interchanges added (optional, only reason to include it is if any users need to use one of the new exits in their list files, and were waiting for us to add them to the route file):

2016-01-23;(USA) California;I-15;ca.i015;Added waypoints for new exits 62, 117, 140, and 148


jteresco:
Thanks.  Not bothering putting in the updates entry, unless anyone wants to push for it.

PR #375, opened 2016-01-26
sync

si404:
a couple of datacheck fixes, some files not yet part of systems...


jteresco:
Likely to have more forthcoming today, or should I pull this in and run an update right away?

si404:
I'm going out in a bit, so no.

PR #382, opened 2016-01-30
Some WV fixes

rickmastfan67:
Extension of WV-44 down to just South of the King Coal Highway till WV gets their act together and only has one posted route for US-52 in that area.  Also a small extension of WV-817 along US-35.

2016-01-30;(USA) West Virginia;WV 44;wv.wv044;Extended southward from it's former southern end at US 52 down to just south of the intersection with the King Coal Highway at Beech Creek Road.
2016-01-30;(USA) West Virginia;WV 817;wv.wv817;Extended westward along US 35 near Henderson from the old western end at US 35 (US35_HenS) to an intersection with Old US 35 (US35_HenN).


yakra:
WV817: It's already US35 at the new US35_HenN endpoint -- a better label would be the cross street *at which* US35 continues...
>For sudden ends at no particular intersection or landmark, the name of the continuing highway can be used if it begins where the highway in question ends, i.e., is not concurrent.

rickmastfan67:
However, the might be extended (again...) along the old route of US-35 into downtown Henderson.  It was supposedly posted as such for a short period of time, and might again.

yakra:
Blarg. Works for me.

PR #385, opened 2016-01-30
CA interstates cleanup part 4A

ovoss:
Error fix on just-updated I-305 route file, based on what turned out to be an out-of-sync version of the I-80BLSac route file. Double-checked to make sure this one will sync with the master versions of I-80BLSac and US 50.


yakra:
Shall I merge this in straight away, or do you think you may have more commits to add?

ovoss:
Eric:

Please go ahead and merge it now. It prevents a broken concurrency for other 
users. Also, the next commit (starting with updating the very large CA I-5 
route file) will take a few days, so I'd prefer to do that separately.

Oscar

----- Original Message -----

PR #386, opened 2016-01-30
usaky3 10XX fixes

mapcat:


yakra:
Looks like KY330 will need coords tweaked to match KY1054. Nothing that should prevent this from being merged in, which I'll probably do soon. Unless you may have more commits coming and would prefer that I hold off.

yakra:
FWIW, Bing seems to be pretty good about matching the proper road names where you've changed the labels.

mapcat:
Thanks for catching KY330.

For KY1070, I checked property listings on the Hickman County assessor's site and the addresses for the KY1070 part of the road are XXX State Route 1070 but the rest are XXX Howell Rd. Breathitt County does not have the equivalent resource, but I was only able to find addresses for houses on Hunting Creek Rd south of where KY1094 supposedly ends.

There were other instances of "End" in this group as well, and if I couldn't find anything that led me to think that the roads' names changed, I left it alone.

Not sure what your Bing comment refers to.

mapcat:
Oh, and I did not change KY1020 and KY1059 yet since I'd like to hear suggestions on what to do with them. Relevant post in the forum: http://tm.teresco.org/forum/index.php?topic=26.msg198#msg198

KY1020: not marked well in Louisville. Trust the routing on the KYTC map?
KY1059: extend to KY32 or not? KYTC says we have the correct end, but GMSV shows a KY1059 sign at KY32 on the supposedly non-highway segment. https://goo.gl/maps/vQ8qRzeGAuu

yakra:
Re: End vs. whatever else:
Deeyom, son! You win the Internet! :D

> Not sure what your Bing comment refers to.

Viewing the Bing map tiles in the CHM Waypoints Editor.
This goes back to the "OSM vs. KYTC / GMSV" topic I remember discussing a bit in another thread.
I had a look around a few of the places you relabeled points in WPTedit, using the Bing tilemap view. Bing consistent labeled the roads as something that'd match your new label, and I presume you saw those same names when looking around in GMSV.

jteresco:
Not sure why the conflicts, but I'll see if I can take care of this now.

jteresco:
And I can!  Will run an update with these changes in a few minutes.

yakra:
Not sure how/why  Update ronreahwy.wpt 331532d got in there.
But now the coords at I-71 mismatch the coords in the I-81 & OH126 files.

jteresco:
I'll fix ronreahwy.

jteresco:
Fixed, but after update that just completed.  I'm sure I'll run another update tomorrow.

mapcat:
The ronreahwy thing was a mistake; I was experimenting with something and didn't realize it would automatically add it to the pull request. Thanks for reversing it.

PR #387, opened 2016-02-01
Merge UT007 and UT007 San

NickCPDX:
The two routes are now connected.


NickCPDX:
Awesome, thank you!

yakra:
WarValRd is in use and should be retailed as an AltLabel for exit 10.

yakra:
WarValRd is in use and should be retained as an AltLabel for exit 10.

NickCPDX:
Not sure how to do that?

mapcat:
Add +WarValRd after the waypoint label (10).

jteresco:
I'll add that also, since I have changes in progress here.

PR #388, opened 2016-02-01
Delete UT 007San

NickCPDX:
Merged to Mainline.


yakra:
You'll also need to delete this from the CSVs, and add UT7San as an AltRouteName for UT7.

jteresco:
I'll merge these in and make the needed changes to the CSVs.

PR #389, opened 2016-02-01
FL: FL-9B

rickmastfan67:
Some minor fixes along this route.  Plus finally adding it's file to the site.

Now, if people think we need to add a 'name' code to FL-9B, I can do so before it's pulled in.  Otherwise, I think it's ok as-is.

Anyways, all of the other changes in this don't need an update entry as all they were was re-centering points based of newer OSM locations that synced with Bing/Google.

Also, as to why I didn't add an exit number for US-1 in both FL-9B's and I-795's file, it's because FDOT might have accidentally flipped the newly posted number (2) it with a new exit that will be opening when 9B is extended down to I-95 this year (Durbin Road, posted 3).  This way, we don't get any future list problems if FDOT does fix the exit numbers there since US-1 is an open interchange.  See this following thread that deals with the exit numbers: http://tm.teresco.org/forum/index.php?topic=27.msg258#msg258


yakra:
Exit 4 is hidden?

yakra:
WRT the name code, I'd say to just be consistent with what you've been doing in usafl so far.

rickmastfan67:
Yes, 4 is hidden as that exit isn't opened yet.

PR #390, opened 2016-02-01
AL: Some I-85 fixes

rickmastfan67:
Just added new exits 14 (opened 01/26/16) & 50 (http://clinched.s2.bizhat.com/viewtopic.php?t=2209&mforum=clinched).  I also re-centered 16 which was way off of the correct location.

No need for an update entry for this one.


rickmastfan67:
Also see some more background on Exit 14 here: http://tm.teresco.org/forum/index.php?topic=61.0

PR #391, opened 2016-02-01
TN: US-27 overhaul

rickmastfan67:
Couldn't sleep.  So, I finally knocked out the overhaul of US-27.  This leaves me only needing to overhaul US-64 up to current standards for TN.  This one has been really lacking since the reroute below happened in 2007 I think, a ~year after I first created the US-27 file (:yikes:).  I'm honestly amazed nobody posted a thread about that one since there was a point on the old alignment there.

2016-02-01;(USA) Tennessee;US 27;tn.us027;Removed from old two-lane through Evensville (now Medlin Pike), and moved onto a new 4-lane alignment to the west between the intersections with Medlin Pike (MedPike_S & MedPike_N).
2016-02-01;(USA) Tennessee;US 27 Business (Dayton);;Route deleted

=====

Also, for the other deleted bannered route files seen here, I have no idea how they got uploaded, as I told Tim to delete them a long time ago (entries in csv files had already been removed, just not the real files).  So, a little cleanup here removing files that shouldn't be here anymore.


rickmastfan67:
Crap, I noticed a minor typo in my update entry.  Meant to say 'east' instead of 'west'.  Jim, I'm going to update the updates file right now with that minor fix.  Just make sure to update your own copy first before you add any more update entries to it to prevent a conflict.

PR #399, opened 2016-02-04
Update ut.ut007.wpt

NickCPDX:
Added a shaping point, plus re-added original waypoint names back into system(did I do that right?)


yakra:
The new-old point labels were not in use (only WarValRd (added in a previous commit) was), and are not needed, I believe.
+X514112 was needed; the centerline was outside the tolerance lines there
+X526413 was not needed; route within tolerance with only +X514112 added
On that note, existing +X441971 is extraneous.

Of course, there is no harm in having any of this stuff in there. If someone else merges this PR, that's fine by me. The greater good = the addition of +X514112.

jteresco:
We can merge this in later today unless there are strong objections to the additional hidden labels (near-zero) cost and a couple unneeded shaping points (minor costs in site update time, DB size, web rendering).

yakra:
I don't really object.

PR #398, opened 2016-02-04
CA Interstates cleanup part 6

ovoss:
Completes updates for CA Interstates, including I-5 and most northern California Interstates. Some affected US and in-dev state routes also modified.

If no glitches, this should complete work on issue #30, with all CA routes in active systems fully updated and synched.

PR #402, opened 2016-02-05
CA I-305 resynch

ovoss:
Resynchs CA I-305 with concurrent routes I-80BL Sacramento and US 50, and intersecting route I-5. This resolves conflicts problem with pull request #398.

### DIFF
--- a/hwy_data/CA/usai/ca.i305.wpt
+++ b/hwy_data/CA/usai/ca.i305.wpt
@@ -1,11 +1,11 @@
-1A http://www.openstreetmap.org/?lat=38.575783&lon=-121.563077
-1B http://www.openstreetmap.org/?lat=38.575035&lon=-121.548722
+1A http://www.openstreetmap.org/?lat=38.575816&lon=-121.563077
+1B http://www.openstreetmap.org/?lat=38.575027&lon=-121.548642
 3 http://www.openstreetmap.org/?lat=38.577091&lon=-121.526556
 3A http://www.openstreetmap.org/?lat=38.573861&lon=-121.518788
 4A http://www.openstreetmap.org/?lat=38.568367&lon=-121.510838
-4B http://www.openstreetmap.org/?lat=38.567268&lon=-121.506515
+4B http://www.openstreetmap.org/?lat=38.56723&lon=-121.506553
 5 http://www.openstreetmap.org/?lat=38.563435&lon=-121.492363
-6 +6A http://www.openstreetmap.org/?lat=38.558771&lon=-121.473556
+6 +6A http://www.openstreetmap.org/?lat=38.558754&lon=-121.473545
 7A http://www.openstreetmap.org/?lat=38.567764&lon=-121.469629
 7B http://www.openstreetmap.org/?lat=38.574515&lon=-121.466689
 7C http://www.openstreetmap.org/?lat=38.578752&lon=-121.464854


ovoss:
Weird that we keep getting conflicts with I-305.

Here's the revised file as submitted. However the conflict is resolved, the file going into the master should look EXACTLY like this:

1A http://www.openstreetmap.org/?lat=38.575816&lon=-121.563077
1B http://www.openstreetmap.org/?lat=38.575027&lon=-121.548642
3 http://www.openstreetmap.org/?lat=38.577091&lon=-121.526556
3A http://www.openstreetmap.org/?lat=38.573861&lon=-121.518788
4A http://www.openstreetmap.org/?lat=38.568367&lon=-121.510838
4B http://www.openstreetmap.org/?lat=38.56723&lon=-121.506553
5 http://www.openstreetmap.org/?lat=38.563435&lon=-121.492363
6 +6A http://www.openstreetmap.org/?lat=38.558754&lon=-121.473545
7A http://www.openstreetmap.org/?lat=38.567764&lon=-121.469629
7B http://www.openstreetmap.org/?lat=38.574515&lon=-121.466689
7C http://www.openstreetmap.org/?lat=38.578752&lon=-121.464854

jteresco:
OK, instead of messing with the conflict again, I'm just going to make this change right in my copy and commit it.  Seems very odd that the copy I get has a +999 alt label on 7C and yours has it in neither the pre-commit or post-commit version.

ovoss:
+999 should probably be added as an alternate label to 7C (it's a label in 
use in CHM, but not in TM). I'm not sure how it got lost in my previous 
edits to the I-305 route file.

----- Original Message -----

jteresco:
How about the +0 for the first point?  I noticed that was also in the previous version.

ovoss:
Yes, same situation as with +999, in use in CHM but not TM.

I took a look at the "CHM final" copy of ca.i305.wpt. It has neither the +0 
nor the +999 labels. But then again, CHM's HB shows neither label as a 
deprecated label for CA I-305 (still, at least one CHM user had those labels 
in a list file). I'm not sure what to make of that.

----- Original Message -----

jteresco:
No harm putting it back in, so I will.

yakra:
**History:**
Looks like things first got messed up between steps 3 & 4 below. (Maybe?)
I wonder why there was no conflict reported at that time. Or maybe it just got merged in with one...

1.) a8cb5c3 California ("TM Original". Matches "CHM final".)
2.) c3a9407 CA I-cleanup 4 -- AltLabels +0 & +999 added
3.) f97f21d CA I-cleanup 4A -- coords changed: 1A 1B 4A 4B 6
4.) 99526cb CA I-cleanup 6 -- **TM Original**, with only Exit 4A coords changed (different from I-cleanup 4A as well)

5.) de1da3d Jim's resolution -- AltLabels put back. All coords as in I-cleanup 4A, except for Exits 4A & 4B which have the coords from I-cleanup 6.

6.) a32c16e What Oscar Posted upthread.
vice I-cleanup 4A: AltLabels missing. Exit 4A has coords from I-cleanup 6.
vice Jim's resolution: AltLabels missing. Exit 4B has coords from I-cleanup 4A.

7.) 8d4c582 AltLabels restored. Otherwise identical to What Oscar Posted.

From my point of view of only looking at the .WPT file code in the various commits (not at the HB or intersecting routes), this looks like the way we'd want things to be. (I think!) Noting however that this may still be a work in progress, requiring 4A & 4B to be resynced on intersecting/concurrent routes as Oscar noted in #398.

*As for CHM users using 0/999, I would just chalk that up as a simple brainfart, coming from the repetition of entering in 0 & 999 for almost all the other clinched routes.*

--

Wow. I've spent far more time looking into this abyss than I really should have...

PR #403, opened 2016-02-06
Kentucky 11xx fixes; add KY361 and US50BusBel

mapcat:
Newsworthy:

2016-02-06;(USA) Kentucky;KY361;ky.ky0361;New Route
2016-02-06;(USA) Ohio;US 50 Business (Belpre);oh.us050busbel;New Route


yakra:
Gave a quick check over some-but-not-all of the files. Looks OK to me.

One thing I'd suggest is on ky.us031wbypeli, US31W_N -> US31W, and checking GMSV if necessary to see whether it's an implied loop, or just connects to the parent at one end.

That's not something that should prevent this from being merged in.

mapcat:
I needed to make some other adjustments (forgot some alt labels) and agree that US31W_N can be just US31W. Heading south, the bypass is only signed as such as far as the Western Kentucky Pkwy, and north the "Bypass" signage begins at the WKP exit. However, in GMSV some "Truck US 31W" signs appear northbound along US31W approaching the WKP, and follow the WKP (along with some Truck US62 signs). Not sure if these should be understood to mean Bypass US31W, or if they're something new, or if they're in Streetview images that are old enough (2011) to be ignored.

mapcat:
Well what do you know. There's a US31WTrkEli in the HB already, which works to complete the loop.

A check of KYTC photologs of the WKP from 2014 and 2015 shows that the Truck US31W signs are still on the WKP and US31W, and have been joined by at least Bus US31W sign on the WKP.

So...leave it as is (two separate files) or combine into some sort of [hybrid bus-truck combo](http://www.hankstruckpictures.com/pix/trucks/mark_wayman/2005/jan09/fun_stuff00/file0012.jpg)?

PR #410, opened 2016-02-10
Activate usamt

si404:


jteresco:
Looking for general comment: do we want to include new devel systems in the system updates list?  I am not sure we want end users doing much with devel systems, and it might be cluttering the list for the preview and active entries that the average user would be more interested in.

si404:
I'm very happy to not put devel systems in there. It would also be nice to have them as opt-in on the Highway Browser rather than clutter it.

si404:
I've made a pull request removing the devel entries https://github.com/TravelMapping/HighwayData/pull/411

michihdeu:
I absolutely agree that devel systems should not be in the list. I also think that end users should not get in touch with devel systems per default (I think we've already discussed about it on the "old" forum". I'll add an issue that devel systems should not be displayed (per default) in the HB.

yakra:
I agree; let's keep devel systems out of systemupdates.csv

michihdeu:
@yakra, if we show them in the HB, what should "normal end user" do with them? I think Jim has suggested on the AAroads forum that they should not be displayed there.
I don't like that new systems directly appear in the HB without notification in the system update list to avoid confusing "normal end user".

si404:
I've removed the devel lines from systemsupdate.csv

PR #451, opened 2016-02-25
In dev changes

si404:


yakra:
Looks OK except for MT US2 WindRdgRanRd = nonstandard label.

>If the cross road name has more than 3 words, use one of two options:
1. Pick out the two most important words besides the road type and use only those: Martin Luther King Boulevard becomes MarKingBlvd. Three words in total are included in shortened form.
2. Pick out one important word besides the road type and use it and the initials of the other words: Martin Luther King Boulevard becomes MLKingBlvd. Two words in total are included in shortened form along with initials of the rest.

jteresco:
Bringing this in now, can fix the label later.

PR #467, opened 2016-03-06
datacheck fixes and devel drafts

si404:


jteresco:
estpt or estt?  The files are estt.csv and estt_con.csv but entries all show estpt.   I'm going with estt to get the update to run since that's also the directory name.

jteresco:
Removing 

mdam;M05;;;mda.m005

as it has no entry in the _con.csv or file.

jteresco:
Removing armm's M5 entries from the csv files since its wpt is invalid (just one point).

jteresco:
Also noticed some misplaced files in BRB but not in a system subdirectory.

jteresco:
Also fixing estt's files to have est.txxx instead of est.ptxxx, another hint that I chose correctly a few comments up.  Leaving route names as "Pt" though - might want to verify that what I did is correct.

PR #483, opened 2016-03-21
Misc CA and NM updates

ovoss:
Some minor point relabels and synchs to active NM routes. Also adds four new routes to in-dev usaca system, changes to existing related routes (one of them active), and a few other updates of existing in-dev routes.  These are just the start of my updating of usaca to reflect my field research in CA, now that I'm back home.


yakra:
CA11: Did you mean for iB to be 1B?

ovoss:
Eric: 

Yes, it should be 1B. But it's not an active route, and it would be a pain to modify (which I don't know how to do) or revert a pull request just for that. I would just let that go for now, and fix it in a few days as I start updating more CA in-dev routes. 

The only reason I included CA 11 in the pull request is to clear my queue of changes I drafted on the road, resynch my copy of the master, and start with a clean slate when I resume route file updates. 

Oscar 

----- Original Message -----

yakra:
Agreed. No need to change it now, especially if it's inactive. You can add to this PR, or just get it in the future. Just making sure you know about it.

Also: The 905/11/125 interchange, once all is said and done, may be a a candidate for "MD I-97 7" style waypoint placement. This was one of the canonical examples from [manual_points.php](http://cmap.m-plex.com/tools/manual_points.php), and had an illustration included before Google discontinued the Staticmap API v1.

PR #484, opened 2016-03-21
usaky3 -> usaky

mapcat:
Leaving the usaky3 directory alone until I'm sure that all the files were copied into usaky


jteresco:
I now see 1451 wpt files in hwy_data/KY/usaky, so this should be good to go.  I've got a few minutes here so I'm going to try an update with these three most recent pulls.

jteresco:
I'm going to blast away the usaky3 directory from the command line.  It will still have history in case we need to refer back to old files there.

PR #486, opened 2016-03-24
A few more CA in-dev changes

ovoss:
Fixes CA 11 typo; truncates CA 14U and CA 54; redoes US 6 Historic to reflect CA 14U truncation and other streamlining; replaces CA 108 Business (Sonora) placeholder with a real route file; (conforming changes to CA 49); and redoes CA 138, CA 177, and CA 14 Truck (which is also signed as I-5 Truck, but less headache to have it in the HB only as a state route).


yakra:
**ca.us006hissan.wpt:**
I-5Trk/CA14Trk isn't a standard label; at the very least, the "CA" would have to go (I-5Trk/14Trk).

This leads into a bit of a sticky situation...
The idea of multiple routes + bannered route(s) in a waypoint has always bothered me. I've typed out some of my thoughts on the subject: tm.teresco.org/forum/index.php?topic=151

ovoss:
I did the same thing for the CA 14U route file.

I'd be OK with just using one bannered route, CA14Trk, in the labels for 
both files, in place of I-5Trk (what's there now) or I-5Trk/CA14Trk (what my 
pull request would change it to). On one hand, the overhead signage for the 
truck route (in my pull request as ca.ca014trknew) uses I-5 and CA 14 with 
equal prominence. On the other hand, I-5Trk currently has no route file, and 
I'm unsure FHWA even recognizes it as a separate Interstate route. (Caltrans 
treats it as a separate route, 5S -- there is no 14S and 5S is not signed as 
such -- since unlike truck lanes elsewhere on California Interstates, the 
truck route follows an independent alignment from the I-5 mainline). Using 
CA14Trk, and ignoring I-5Trk, would best fit our having only a CA14TrkNew 
route file.

----- Original Message -----

ovoss:
I'm now inclined to go with I-5/14 as the labels for US6HisSan and CA 14U, 
and make that change in my next pull request. You need to go briefly on the 
truck lanes to get between the Sierra Hwy. and the I-5 and CA 14 mainlines, 
but for simplicity's sake I'd ignore that. Note that their interchange with 
the truck lanes does not match points in either the I-5 and CA 14 route 
files, so any intersecting-route detection won't pick it up -- but then, the 
same is true for double-trumpet connections between some eastern toll roads 
and connecting freeways, which is OK per the point placement manual.

I might also delete ca.ca014trknew, treating that as just separate 
truck-lane roadways and ramps within the complex I-5/I-210/CA 14 Newhall 
interchange clusterfunk, rather than as a separate route. While I-5 and CA 
14 are exit-numbered, there is no separate exit number for the truck 
lanes/Sierra Hwy. interchange.

----- Original Message -----

PR #487, opened 2016-03-24
fix errors, and A6182

si404:
2016-03-24;(UK) England;A6182;eng.a6182;Route extended at south end from M18 to A638


michihdeu:
Just my 2 cents:
brbabc;BRB;Barbados ABC Highway;**brown;4**;devel
brbh;BRB;Barbados Highways;**brown;4**;devel
Why are both tier 4?

I think you should add to systemupdates.csv:
**2016-01-01;Czechia;czed;Czechia Dálnice;extended**

2015-08-23;Europe;eursf;Europe Select Named Motorways;active
**2015-08-23;Europe;eursf;Europe Select Named Motorways;extended**

PR #490, opened 2016-03-27
asiaho fixes

si404:


jteresco:
OK, will try this tonight.  Seem to be a lot of list file updates coming today, so I'll give that a few more hours.

PR #492, opened 2016-03-28
Jamaica T3 completion

si404:
2016-03-28;Jamaica;T3;jam.t003;Route extended at north end from temporary terminus with A1 at Monleague to the A3.


jteresco:
For some reason this came to me with a (I believe meaningless) conflict.  I resolved it.  Please verify that jam.t003 is ok.

PR #497, opened 2016-03-29
asian highways

si404:
fixes concurrency errors with eure
changes outlines in NE China, Mongolia and West China to actual routes.


jteresco:
Removing the first root from this line in asiaho_con.csv, as there is no corresponding entry in asiaho.csv.

asiaho;AH42;;;chngs.ah042,chnqh.ah042,chnxz.ah042,npl.ah042,ind.ah042

jteresco:
I'm commenting out asiah and asiaho to get a site update to run.  I found and fixed one misplaced file (was still in an asiaho directory, should be in asiah).  The next one I found 

FileNotFoundError: [Errno 2] No such file or directory: '../../../HighwayData/hwy_data/HKG/asiah/hkg.ah368.wpt'

I decided to go this way to get the site update through pending fixes for this and a run through checking for others.  The problem is that it's about 3 minutes into the site update when such errors are found...

PR #520, opened 2016-04-16
CANBC point request

ovoss:
Tweaks bc.bc023nak route file, to replace a shaping point between HalRd and BC 31, with one better reflecting my travels on the route. I traveled south from the ferry terminal, past the Central Kootenay Regional District line, to a now-closed private road to the old Halcyon Hot Springs -- exiting the highway about a mile north of the HalRd waypoint.

Si, is this OK with you?


si404:
This is fine with me.

PR #521, opened 2016-04-16
More China

si404:


jteresco:
Just a few this time, but some that I think you'll need to resolve:

```
Could not find Route matching root chnzj.g25lis in system chng.
Could not find Route matching root chnha.g40 in system chng.
Could not find Route matching root chnjx.g6001 in system chng.
No roots in _con.csv line [chng;G6001;;;chnjx.g6001]
```

PR #522, opened 2016-04-17
chng fixes

si404:
fixes errors in #521


jteresco:
Looks like this one will go through, thanks.

PR #523, opened 2016-04-17
Update al.us431.wpt

ajfroggie:
US 431 relocated onto the Anniston Eastern Bypass and a 3-mile leg of I-20.


jteresco:
Will be putting in an updates entry, please make sure it looks ok.

PR #524, opened 2016-04-17
Update al.us078.wpt

ajfroggie:
Point name correction from removal of the US 78/US 431 concurrency in/near Anniston.


jteresco:
Assuming no updates entry needed for this one.

ajfroggie:
Nope.  Just changed two point labels to account for the US 431 realignment.  Old point labels remaining as hidden labels in case anyone was using those points.

PR #530, opened 2016-04-20
india and china expressways -> preview

si404:


jteresco:
Rerunning just the "TravelMappingDev" DB update now with this change.  Will hold off on a regular update pending any further data or list file updates before I call it a night.

PR #539, opened 2016-05-01
MA2ACam: restore MA2_Lin AltLabel

yakra:


yakra:
Fixes for the [PA: PA 29 and PA 100-minor Lehigh Valley errors](http://tm.teresco.org/forum/index.php?topic=202) forum thread are along for the ride.

PR #543, opened 2016-05-02
deub drafts from Michael

si404:


jteresco:
Commenting out deub for now - the csv files are in ISO-8859 text but need to be UTF-8 Unicode text to be readable by my current Python program.

I can fix this at some point, but time is very limited this week.

michihdeu:
Ok, my bad. I'll save it UTF-8. Don't upload them because some file names were not according to the rules...

PR #549, opened 2016-05-07
Georgia State Highways

theFXexpert:
Adding usaga files.
- 110 / 471 .wpts now drafted. Contains mostly bannered routes along with some higher numbered routes so far.
- Commented out of systems.csv for now.


jteresco:
You can uncomment usaga's entry in systems.csv if the usaga.csv and usaga_con.csv files only have the entries included for routes plotted so far.  That would allow you to see them in the highway browser (though not in the other maps until the system can go to preview).

theFXexpert:
Yeah. I wasn't sure if a site update would run properly if wpts were missing. #550 has csvs that only include existing wpts so far.

PR #558, opened 2016-05-10
Begin usaky5

mapcat:
Adds usaky5 to systems.csv


jteresco:
Going to be commenting out usaky5 from systems.csv since there are no csv files committed.

PR #570, opened 2016-05-16
More deub files emailed to me.

si404:


jteresco:
Michael, please be sure to get the csv files I have updated before the next round of edits.  I removed extraneous ctrl-M characters at the end of the lines, lots of semicolons that didn't belong, and some blank lines that are not allowed in files.

jteresco:
Also removing 

deub;B10;;Stuttgart;deu.b010

since there is no .wpt

michihdeu:
Sorry! I forgot deu.b010 and my csvs contained some blank lines, my bad.

I've compared the uploaded csv files and my offline files. I think they are absolutely identical, no additional semicolons! And ctrl-M characters... I guess you mean CR LF, yes, they are in my files. I edit with Notepad++, encoding UTF-8 and "Show all characters" option enabled. I could replace "\r\n" when I'm done, all entries would be in one line but it's not in one line on github!?!?!?! I made an attempt with open office right now but it's the same result.... Still CR LF there. How should I handle/edit csv files?

For any reason, my latest update broke my list file: "Waypoint label(s) not found in line: DEU E52 FRA/DEU 54(A5)". But it's still in the wpt file!?!?!?!?!?

:-(

michihdeu:
And all my damn files seem to contain bullshit again "LABEL_INVALID_CHAR"........... Like [last time](https://github.com/TravelMapping/HighwayData/pull/562) and I've no idea.........

:-( :-(

mapcat:
The easiest way I've found to avoid formatting errors is to edit directly in the GitHub web app.

To make .csvs, I use Google Sheets. Make columns for each value in the string, and then create the string in the final column using =CONCATENATE . No problems so far.

What's your email? I can share my file with you if you want to see how, or create one for you if you need me to.

michihdeu:
Thanks, I don't use google sheets, have no google account. I still hope that it's possible with "standard" tools like Notepad++.........

mapcat:
You could do it with any spreadsheet...I've not used OpenOffice recently but I suspect it also has the concatenate function.

jteresco:
Sorry I can't be too much help here, since I don't use Windows.  But a couple things that I think might help.

Before making any edits, always start with a version of the files downloaded from GitHub.  Even if you don't want or are unable to install any git or GitHub software on your own machine, you can always download the repository as a zip file.  Ideally everyone will adopt what most contributors are doing and create and edit files within a git repository and submit changes by pull request.

When editing .csv files, avoid blank lines, make sure you have the right number of fields, and use either plain ASCII text when possible or UTF-8 encoding when international characters are needed.  All .wpt files are required to be plain ASCII text, since international characters are not allowed in those files.

Also with csv files, most spreadsheets, including Excel and OpenOffice Calc have the option to load and save csv files.  If you set the delimiter to ';' it should put the fields in cells and allow you to edit, and will save correctly.  I don't do this, though.  I prefer to edit all in a regular text editor.

jteresco:
I verified that many of the deub .wpt files are in a UTF-8 encoding.  I think I can pretty easily convert.  Once I've done that, I'll push the changes up to the master and then you can work from those.

si404:
I use notepad to edit all TM files. The return key works perfectly fine for making new lines in both the file you have and in github (and should do in Notepad++)

michihdeu:
Thank you very much for helping! I have to check the options when I have more time... Could be next week though. I'll also check mapcat's csv option.

Some notes:
> I verified that many of the deub .wpt files are in a UTF-8 encoding

I thought UTF-8 encoding is required for wpt files. What's the required encoding?

I guess my wpt files have a general problem. I've modified DEU A8Ros and DEU E52 which are entered in my list file. The [log file](http://tm.teresco.org/devlogs/michih.log) reports now that the waypoint labels are unknown. Could it be caused by UTF-8 encoding?

> And all my damn files seem to contain bullshit again "LABEL_INVALID_CHAR"........... Like last time and I've no idea.........

It's fixed for all files except deu.b006 and deu.b010 now. Was it caused by UTF8-encoding and fixed by Jim?

jteresco:
.wpt files (and .list files) are required to be plain ASCII.  This is a carry-over from CHM, but I think still good policy.  

I converted all of the files I had in deub as of a couple days ago to ASCII and committed them to the repository.  A few UTF-8 .wpt files have crept back in or slipped through:

./DEU/deua/deu.a008ros.wpt
./DEU/deua/deu.a094mun.wpt
./DEU/deub/deu.b006.wpt
./DEU/deub/deu.b010.wpt
./DEU/deub/deu.b013.wpt
./DEU/eure/deu.e45.wpt
./DEU/eure/deu.e52.wpt
./DEU/eure/deu.e53.wpt
./DEU/eure/deu.e552.wpt

I can convert them to ASCII and commit unless there is an objection.

michihdeu:
Ok, thanks!

Issue #571, opened 2016-05-16
IN: IN267 truncation and a few more tweaks


ajfarina:
1) IN267 (Plainfield) has been decommissioned north of US40 and the concurrency has been removed. I was out there today and drove around Plainfield to confirm the changes. New signage is in place too now.

2) With the changes to IN267, the points on US40 need to be tweaked in that area. Also, the labels "IN267" on US36 and US136 hadn't been updated yet.

3) While on US36, might it be possible to add a point at Ronald Regan Parkway (or move the Raceway Road point a half mile west)? That would be a much more fitting location.

4) The point on US31 "IN44" needs to be changed, since the route has been removed from Franklin.

Thanks!

ajfarina:
I forgot to include another:

5) IN38 (Noblesville) follows 2nd Street eastward to Main, then southward to 6th, then eastward to White. It does not follow 2nd to White.

6) Apparently IN252 is being decommissioned between I-65 and IN9 once reconstruction wraps up. I'll keep you posted on when that becomes official.

Thanks.

ajfarina:
Just saw this one tonight: There is a new IN9Trk in Shelbyville to be added. (Yes, it is a real one with real shields, etc.) It follows IN44 and I-74 to and from IN9. It allows trucks to bypass the circle at the center of town and the low/narrow railroad crossing just north of there.

Thanks.

PR #572, opened 2016-05-16
Deal with issue #571

si404:


jteresco:
I think this pull broke many lines in datacheck.csv.  I think it's simply that trailing zeros were removed from the last field. @si404 could you revert those changes?  My datacheck false positive checker is a simple pattern match and it doesn't see "11.6" as being the same as "11.60".

PR #577, opened 2016-05-18
Fix datacheckfps.csv

si404:
Put back the trailing zeros - that's what you get for using a spreadsheet to re-alphabetise it.


jteresco:
I figured it was something like that.  Thanks.

PR #585, opened 2016-05-23
Final usaky8 cleanup

mapcat:
Looks ready to be moved to preview.


jteresco:
So is this one something I should merge in, or was it duplicated, and hence closed without a merge?

PR #586, opened 2016-05-23
Final usaky8 cleanup

mapcat:
Looks ready to be moved to preview.


jteresco:
Never mind my comment in #585, I see this is the same stuff.  Will merge in and be included in tonight's site update.

PR #590, opened 2016-05-26
gbrtr

si404:


jteresco:
Commenting out gbrtr from systems.csv, as it appears there are some missing, misplaced, or misnamed wpt files.  First error:

`FileNotFoundError: [Errno 2] No such file or directory: '../../../HighwayData/hwy_data/SCT/gbrtr/sct.angcoatr.wpt'`

It looks like there are 16 SCT routes in gbrtr.csv, but only 4 wpt files in the directory.

Also looks like this new system was sent directly to preview, but should probably stay devel until it at least has been given a quick look in the HB, right?

Going to commit a couple quick fixes I made before I hit this roadblock and will retry the update.

si404:
It's an expansion of the Scottish Tourist Routes system to other parts of the UK (and adding some more Scottish routes), which was preview. I'll make a pull request that fixes it.

PR #593, opened 2016-05-26
DEU B30-B39

si404:


michihdeu:
Was there any problem with my files? I've downloaded them from Github first. If there was no problem, I'll proceed this way. Thanks for your support and patience!

jteresco:
No problems on my end.  Looks like a good procedure.

PR #594, opened 2016-05-26
Fix UK Tourist Routes

si404:


jteresco:
I don't have a strong opinion, but should the "valley" abbreviations be "val" instead of "vly" in both route names for .lists and file names for .wpts?

jteresco:
So far just minor fixes needed.  Seeing if I can get the update through now.

yakra:
I'd prefer Val over Vly. Just my 2c.

si404:
OK, I'll get around to changing that some time soon.

jteresco:
Thanks.  No rush as far as I'm concerned.

PR #595, opened 2016-05-28
deub 40-72

si404:


jteresco:
I'm not going to merge this in right away, since it looks like almost every (maybe every?) line in the previously existing files was changed. I just want to make sure we don't have file encoding issues or anything like that.

jteresco:
It also looks like we have some wpt file names with capital letters in them, which I don't think we've done before.  It probably wouldn't cause problems, but I think it would be better to keep all lowercase unless there's a good reason in this system to have the mixed case filenames.  Some of those files might have gotten into the earlier deub submissions, but I think some are new here.

michihdeu:
>  it looks like almost every (maybe every?) line in the previously existing files was changed

138 files were uploaded. I saw a general fault with concurrent numbered roads. I had waypoint labels like "A3(27)" instead of "27(A3)". I've fixed that in some files and made other minor modifications. In total, 25 out of 138 existing deub files have been changed. In addition, I've added 69 new route files (B40 to B72).

B1 to B72 is almost complete now. I'm not sure about B28/B28a routes. It was announced to be changed in Mid 2016 but I couldn't find reliable info. A short B32 segment in Horb is also concerned by the B28/B28a change and still missing.

In total, I think I've mapped about 30-40% of deub system according to the network length right now. Many (shorter) routes are still missing (B73+).

> I just want to make sure we don't have file encoding issues or anything like that.

I've downloaded all files from Github before editing. I don't expect problems because the previous upload already worked fine...

> It also looks like we have some wpt file names with capital letters in them

I've already had these files and it seems that they didn't cause any problem, for instance http://tm.teresco.org/devel/hb.php?r=deu.b009Ast and http://tm.teresco.org/devel/hb.php?r=deu.b032Ast. I've realized right now that previous files with capital letters have been renamed by Si or Jim.

It's because I have to refer to it, for instance: http://tm.teresco.org/devel/hb.php?r=deu.b032. I refer "B32Ast". I cannot call it B32 because it would be a self-ref error (btw: which happened in the first line of the file - should be B27 - which would be fixed with the latest update).

I was about to ask earlier but I had to less experience and didn't know all use-cases. I think I should just use lowercase in file names but capital letters only(!) for refering within the wpt files, like http://tm.teresco.org/devel/hb.php?r=deu.a005ast. That means, "deu.b032ast" but "B32AST" waypoint in "deu.b032". What do you think?

I'll be off for minimum one week now. I think we should discuss the route naming issue when I'm back.

Maybe deub system should be split like usaky system and I could bring a first set to preview status so that an experienced CHM contributor could make a peer-review? Maybe when B1 to B100 is complete? The current deub numbering range is from B1 to B588 (plus B999) with many gaps in-between.

Have a nice week!

si404:
The mass upload function in Github inbrowser (its easier for me to use that, rather than the desktop version when dealing with files emailed in) decides that every file replacement is completely new, rather than making a collapsed change log (though it did it for the .csvs). The actual changes are highlighted in stronger highlight

jteresco:
OK, I'll pull this in for the next site update, which I plan to do very soon.

PR #596, opened 2016-05-29
OpMuxOK part 3

yakra:


jteresco:
Taking care of the I-40 datacheck error that was introduced here.

PR #603, opened 2016-06-04
svng

si404:


jteresco:
svng.csv has an encoding being seen as "Non-ISO extended-ASCII text" and I can't tell what the international characters there are supposed to be.  @si404 could you save as utf-8 on your end and submit an update for that?

jteresco:
Also noticed that the filenames for the routes whose entries in the csv are problematic also have incorrect file names.  I could fix but I'll leave it to you to clean up the whole batch.

jteresco:
Will put back svng - that last commit commenting it out was unintentional.

PR #607, opened 2016-06-08
norsf, hunf and bgrii

si404:


jteresco:
Removed the new US 20 historic segment from the CSVs since the wpt is missing.  Other fixes so far are minor, but no success with a site update just yet.

PR #608, opened 2016-06-09
srbb

si404:


jteresco:
Fixed encoding to utf-8 for hunf csv files.

PR #619, opened 2016-06-17
deub B28/B90-B99 / System split to deub0 + deub1

michihdeu:


michihdeu:
For any reason, the old commits are here again. I've compared to http://tm.teresco.org/forum/index.php?topic=284.0 and I don't know what I've made wrong....

jteresco:
I'm going to have to remove deub0 and deub1 from the next site update.  It looks like the same routes are listed in both, and the deub1 files still use deub0 as their system entry.  Multiple entries for the same route are not allowed, sorry.

As for the old commits staying in, I don't use the same process, so I'm probably not much help...  But we want to resolve it before long, as your pull requests will get larger and larger and have a lot of redundant information.

PR #620, opened 2016-06-17
deub1 fix

michihdeu:


michihdeu:
Sorry, my bad. Hopes it works now! The commit problem seems to be gone but I'm not sure why... ;)

jteresco:
OK, good.  Re-running site update with these two included now.

PR #623, opened 2016-06-18
NC: US-70 Bypass (Goldsboro)

rickmastfan67:
Wasn't included in the system files when added to the site.  This fixes that.
http://tm.teresco.org/forum/index.php?topic=278.msg1455#msg1455


rickmastfan67:
Hey Jim, on a side note, I just noticed that the entry for this on the 'update' page is missing the "()" around Goldsboro in the 'Route' section.  Might want to fix that.

https://github.com/TravelMapping/HighwayData/blob/master/updates.csv#L9

jteresco:
Thanks, fixing.

PR #632, opened 2016-06-22
deub B187-199 + B505 / deub1 split to deub1..5

michihdeu:


jteresco:
Was going to bring this in, but GitHub is showing conflicts, which I will investigate and hopefully resolve tomorrow.

jteresco:
Simpler than I thought.  There were two changes to the same part of systems.csv by you and by @sis404.  No worries.

PR #653, opened 2016-06-29
datacheck

si404:


si404:
2016-07-01;Poland;A1 (Łódź);pol.a001;Route extended at south end from exit 20 to exit 27
2016-07-01;Poland;A1 (Piotrków Trybunalskiego);;Route merged into Łódź section
2016-07-01;Poland;E67;pol.e67;Route removed from DK12 and relocated onto S8 and A1 between DK12/91 and A1 exit 25
2016-07-01;Poland;E75;pol.e75;Route removed from DK14 and DK1 and relocated onto A1 between A1 exit 20 and A1 exit 25
2016-07-01;Poland;S8 (Oleśnica);Route extended at east end from DK12/91 to A1
2016-07-01;Turkey;O5 (Orhangazi);tur.o005orh;Route extended at north end from D130 to O4

PR #656, opened 2016-07-01
datacheck

si404:


yakra:
marking ita.a018sir;SP26;;;VISIBLE_HIDDEN_COLOC;ita.e45cat@+X242320 as an FP?
I think I wanna look into that a bit and form an opinion...

si404:
The A18 ends, with the E45 taking what are sliproads (of what will be a trumpet interchange), but there is no de facto junction there.

yakra:
Aah, gotcha. A little bit similar to I-22, then.

jteresco:
Removing idn.ah002 entry from asiah.csv, no corresponding wpt file.

jteresco:
Also from asiah_con.csv

PR #657, opened 2016-07-01
deub: B2xx completion

michihdeu:


michihdeu:
Dunno what went wrong but it seems I've uploaded the same files twice?!?!

si404:
I don't see that. Seems fine to me.

michihdeu:
Yes, it seems to be ok, I've already deleted the comment.

PR #682, opened 2016-07-15
Misc LA Updates

mwasleski:
Here are a few LA updates that I made before stepping away from the project due to life.  One newsworthy update:

2016-07-15;(USA) Louisiana;I-12;la.i012;Relocated point for exit 1B to proper location at LA3064 and added new point for exit 1C at LA73 and LA1068.

None of the other changes involve any points currently in use.


jteresco:
Thanks.  I think we can even leave out that one updates entry, but if anyone wants to argue in favor of including it, I'm happy to do so.  I'm going to run an update soon that will include these.

PR #687, opened 2016-07-18
deub B450-B498 and minor fixes

michihdeu:


jteresco:
Got an error from a non-ASCII file.  Fixed label in b486 not to have international character.

PR #689, opened 2016-07-19
deub4 to preview; deub0-5 merged to deub!

michihdeu:


michihdeu:
All files have been changed because of the merge. I've checked the number of routes (643) which seems to be correct in the csv files and deub folder. I think it should work although I'm not sure that my "systemupdates.csv" modifications are correct:
- The "extended deub" system is new and the "merged deub4" is also preview now, Maybe an entry for "deub4 to preview" should be added?
- I guess the links to the old deub0..deub5 systems will automatically be detected as not working?

Please check the file and correct it if required!

PR #702, opened 2016-07-27
MT US2 point request

ovoss:
Request to add point in Cut Bank, at east end of detour around downtown segment closed for reconstruction, which is now the only part of US 2 I haven't traveled. The west end of the detour is just east of SR213, no waypoint needed there,

Si, OK with you?


jteresco:
I'm going to assume this is OK and will merge in for the upcoming site update.

PR #708, opened 2016-07-29
US 84 Opp

ajfroggie:
Remove BUSINESS US 84 Opp designation and rename waypoint.


yakra:
https://github.com/TravelMapping/HighwayData/blob/master/hwy_data/AL/usausb/al.us084busopp.wpt

still exists

PR #709, opened 2016-07-29
US 82 Centreville

ajfroggie:
Update US 82 onto the Centreville bypass.


yakra:
This one looks good to merge in.

@ajfroggie, how does this updates text look?
2016-07-30;(USA) Alabama;US 82;al.us082;Removed from AL 25 and AL 382 and relocated onto the northeastern Centreville bypass between AL 25 North and CR 20.

jteresco:
This one is also a file in chm_final instead of hwy_data, so I'm going to revert.

PR #710, opened 2016-07-29
Delete al.us331altopp.wpt

ajfroggie:
Route removed by AASHTO and ALDOT in 2010.


jteresco:
Going to try to back this one out, since it mistakenly removed the file from chm_final instead of hwy_data.

PR #711, opened 2016-07-29
US 331 Opp

ajfroggie:
Remove ALT US 331 Opp designation and revise waypoint.


yakra:
https://github.com/TravelMapping/HighwayData/blob/master/hwy_data/AL/usausb/al.us331altopp.wpt

still exists

PR #728, opened 2016-08-08
DEU eursf

si404:


michihdeu:
:+1:  I think "L70" should be removed from "updates.csv" line:

`2016-08-08;Germany;L70 Kiesbergtunnel Wuppertal;deu.l070;New Route`

PR #730, opened 2016-08-09
Data check unmatched FPs removed / autb csv draft / cheh H17-H30 (completed)

michihdeu:


michihdeu:
Not sure that the file name modification worked...

jteresco:
If it didn't, I can easily fix on FreeBSD next time I run an update from that computer.  Thanks for the fix.

jteresco:
The L to l file renames had not gone through, but they're in now.

PR #736, opened 2016-08-11
CHE routes tweaked / data check

michihdeu:


michihdeu:
@si404 : Would be great if you could check my latest modification (AST banner) too. I absolutely agree that this modification is necessary because the "View Associated Routes" did not work in the HB. Thanks for your advice! I've also added city names to these (very) short segments.

si404:
Looks good

jteresco:
I'm going to pull this in now, since today's site update is an early one.  If there are further problems, submit another and I'll get them in tomorrow.

PR #744, opened 2016-08-13
ND US85BusWil: endpoint label corrections

yakra:


ovoss:
Actually, the old point labels are correct. With US 85 rerouted onto the western bypass of Williston, US85Bus no longer connects at either end to US 85, it connects only to US 2. Nor is there any signage on the rerouted US 85, or on US 2, linking US 85 and US 85Bus over US 2.

It would be logical to decommission US85Bus (which is completely concurrent anyway with US2Bus). But AFAIK that hasn't happened yet, and US85Bus is too well signed in downtown Williston to justify its removal from the HB at this time.

jteresco:
I'll wait to pull this in until @yakra can weigh in and decide if what he put in the request should go in given @ovoss's comments.

yakra:
Brain fart on my part. Oscar's comments make sense. Closing this PR.

PR #746, opened 2016-08-14
deub FPs / swe eursf / autb B1 - B10

michihdeu:


michihdeu:
@si404, I've added some Swedish EURSF routes according to https://sv.wikipedia.org/wiki/Lista_%C3%B6ver_motorv%C3%A4gar_i_Sverige#Motorv.C3.A4gsstr.C3.A4ckor_i_Sverige. They are located in a subfolder and not yet included to the actual csv file (that means, the pull request can be merged anyway, because the new routes will not be visible on the site).

Additional sources:
EnkVag: https://sv.wikipedia.org/wiki/Enk%C3%B6pingsv%C3%A4gen
MalSlaVag: https://sv.wikipedia.org/wiki/Malmsl%C3%A4ttsv%C3%A4gen
VasLed: https://sv.wikipedia.org/wiki/V%C3%A4sterleden,_%C3%96rebro
ArbVag: https://sv.wikipedia.org/wiki/V%C3%A4sterleden,_%C3%96rebro

I think, **SunMot** (Sundsvall motorvägen) should end at exit 232 instead of 236 b/c it's not 2x2 from 232 to 236 and **AutoStr** (Autostradan, Malmo) could be merged with LundLinMot (Lund - Linderöd motorvägen) which had to be renamed to MalLinMot though.

Please check if the routes are correct and add them to the actual csv file for activation.

jteresco:
Best I can tell, I can safely pull this in, so I will.

PR #753, opened 2016-08-16
Update vt.vt025.wpt

ajfroggie:
Replaced point at Chase Hollow Rd with point at Chelsea Rd.


yakra:
Shaping now out of tolerance. Suggest adding ChaHolRd back in. O:-)

PR #758, opened 2016-08-20
autb completed / auta + auts route names

michihdeu:


michihdeu:
@jteresco  There might be a problem with autb because some "B roads" have L prefix. All file names have "b" prefix though. If there's a problem, please rename the names of 11 files starting with "l" instead of "b", see csv file. Thanks.

jteresco:
@michihdeu I'll try them out tonight and rename if needed, thanks.

PR #775, opened 2016-09-01
cze

michihdeu:


michihdeu:
@si404: the Czech eursf candidates are not yet in the csv files but I have some potential wpt files added. eursf is an active system, I think you should check the routes before they get active!

PR #776, opened 2016-09-02
eursf additions checked (or is that czeched?)

si404:
tons of updates already added to updates.csv

also fixed bihm so it can be unhidden and not break anything.


michihdeu:
Thank you very much :+1: 

I guess Swedish L226 does not meet the rules? What's exactly the reason why you've omitted it?

jteresco:
Looks like a big one, so I'm going to merge this in and try a site update now.

michihdeu:
@si404 : I looked for Czech gaps which were not clinchable at all. I see now that there are two "little" gaps in the southeast of Prague which are "only" E roads but should be eursf too: E65 b/n MO and D1 and the eastern MO extension up to D10. I think I should add them, shouldn't I? I would take the names from wikipedia and could activate it directly...

michihdeu:
I've directly added the 2 Czech routes: #778. Swedish L226 is still open.

PR #780, opened 2016-09-04
deu eursf

michihdeu:


michihdeu:
Please refer to commit description for required **update.csv entries** (3 new routes),


@si404 : An additional 8 routes are in the German eursf "TBC" folder. Would be great if you could check them that they really meet the rules. Some comments:

deu.autzubsch is the former A104 which is officially an A100 spur now but it's not signed A100 (https://de.wikipedia.org/wiki/Bundesautobahn_104). Dunno if it should be an eursf route or part of the deua system?

If former A104 is gonna be be a spur, former A102 (https://de.wikipedia.org/wiki/Bundesautobahn_102) and former A105 (https://de.wikipedia.org/wiki/Bundesautobahn_105) - both indicated as "Anschlussstelle" (= interchange) might be spurs too but I don't think they should be eursf routes!?!?

Must eursf routes be (minimum mostly) access-limited, with "happy car" sign? Some of my 8 suggested routes do not fulfill this...

jteresco:
So are these ready to merge and go live?

michihdeu:
Yes.

PR #803, opened 2016-09-18
Datacheck, KY & MI

mapcat:
Added new Exit 10 to ComMtnPkwy and Exit 29 to CumPkwy / I-66FutBow in Kentucky; extended US41 N end in Michigan; fixed concurrency & datacheck

2016-09-18;(USA) Kentucky;KY 974;ky.ky0974;Moved from KY 15 and Old Kiddville Rd onto relocated Kiddville Rd between KY15 and ComMtnPkwy
2016-09-18;(USA) Michigan;US41;mi.us041;Extended north from FortWilSP to ManRd_N


yakra:
> 2016-09-18;(USA) Michigan;US41;mi.us041;Extended north from FortWilSP to ManRd_N
@mapcat, Does this reflect an extension to the actual route, or just a correction to something we had wrong in TM?

mapcat:
Can't recall. It was shortly after I had taken over maintenance of Michigan, and when I checked the end of the route, the existing location didn't make sense. Looking at it after a few more years of doing this, I suppose that the entry could have been phrased more clearly. Assuming the original end point was incorrect, should I go back and edit this to "Moved north end from FortWilSP to correct location at ManRd_N"? FWIW, there are 9 users who have made it to FortWilSP but no further north, so after over 4 years it seems likely that at least a few of those users are aware that the northern end is past where they turned around, rather than people who don't pay attention to the updates.

PR #818, opened 2016-10-01
QC A-73 completion

ovoss:
Extends main A-73 route south to QC 204 in St-Georges, and deletes A-73 (St-Georges), with today's opening of new connection between those route segments.

Updates items:

2016-09-30;(Canada) Quebec:A-73;qc.a073;Route extended south from exit 61 near Beauceville, to QC 204 in St.-Georges, absorbing former A-73 (St-Georges) segment

2016-09-30;(Canada) Quebec;A-73 (St-Georges);(NONE);Route deleted (folded into rest of A-73)


ovoss:
Jim: 

Only the second of my two Updates entries was added to the Updates table. The rest of the A-73 update is OK. 

Oscar 

----- Original Message -----

jteresco:
Found and fixed the problem - there was a colon instead of a semicolon for one of the delimiters in the first.  It should show up on the next update.

PR #819, opened 2016-10-02
More Russia, and norrv activation

si404:


jteresco:
I'm going to go ahead and pull this in, assuming the datachecks are all known to be FPs.  I will not add the FP entries, though.  Please do so if they are.

PR #822, opened 2016-10-07
London Underground Rail Mapping Test

si404:
Putting in a rail system to show concept, will delete/move once separate TravelMapping/RailwayData repository created, or concept shown - whichever is sooner.


jteresco:
I'm not sure how best to accomplish this to start.  I suppose there's really no harm in having it in as a devel system temporarily.

jteresco:
I think I'd rather put this only in the test DB, especially since it's put in as a preview system.  I'll see if I can get it set up.  It's been a while since I was making DB or other code modifications that required the test DB but I think it will still work.

jteresco:
I pulled this in and was able to run a mini site update for the test DB.  From any TM page, add dbname=TravelMappingTest to see things from the test DB.  You have to re-type it as you go from page to page since there is no mechanism in place to carry it along (like we do for u=, for example).  It contains the new rail system and gbna since I am almost sure there are assumptions built in that at least 1 active system exists.  It's built from a temporary new csv railsystems.csv so the good systems.csv can be kept clean and for highway systems only.

There were also 3 csv entries that had no corresponding wpt files, so I removed those.

These updates take only about a minute so I don't mind running them once in a while.

PR #839, opened 2016-10-21
BC TCH 1 changes

ovoss:
Fixes broken concurrences between TCHMai and TCHVan, and corresponding BC 1 routes, by adding points to TCH files.

No Updates items needed.


jteresco:
Was just about to run a site update, so this will go in very soon.

PR #849, opened 2016-10-22
NB145: relocation at E end

yakra:
The the best of my knowledge (though I could be wrong here), the route wasn't actually relocated here, but just incorrectly plotted in the first place.

As this route currently has 0 total drivers, I'm hoping that will allow me to get away with not writing any update text. An unnamed road is involved, and an old endpoint labeled simply "End". It would be like the 2015-09-03 update for PE168, with all its references to lat/long coords. Only worse.
If anyone insists loudly enough, I can write update text.


jteresco:
I'm all in favor of leaving out updates text that we know doesn't affect anyone.

PR #865, opened 2016-10-30
System isls to devel status

Duke87ofST:


jteresco:
A few fixes I'm going to commit and push to the csv.  Missing some semicolons and had some routes in the csvs that didn't have wpts in the right place.  But really pretty good for a new csv!  Also moving the systems.csv entry to alphabetical order within the devel systems.

PR #866, opened 2016-10-30
Update ma.ma110.wpt

ajfroggie:
Add a point in downtown Lowell for a street that ties into the end of the Lowell Connector.


yakra:
I'd merge this in right now, but my antiquated mobile browser says no. Anyone else, feel free.

PR #881, opened 2016-11-02
Review

si404:
2016-11-02;Hungary;E66;hun.e66;Route removed off F801 and F7, and relocated onto F8 between F801 junction and junction with eastwards R7.


michihdeu:
`junction with eastwards R7.`

R7?

PR #887, opened 2016-11-03
OH & IN fixes; datacheckfps for KY & DE

mapcat:
2016-11-02;(USA) Ohio;Atlantic Blvd;oh.atlblvd;Route added
2016-11-02;(USA) Ohio;OH 164;oh.oh164;Route moved from Lincoln Way (US 30) onto Washington St between Lincoln Ave & Market St, and moved from Lincoln Ave onto Market St between Washington St & Lincoln Way
2016-11-02;(USA) Ohio;OH 242;(NONE);Route deleted
2016-11-02;(USA) Ohio;OH 347;oh.oh347;Route extended from US 33 to OH 287
2016-11-02;(USA) Ohio;OH 533;(NONE);Route deleted


yakra:
MD/DE should be the primary label; border labels take precedence over intersecting routes/roads/exit#s/etc where both are in the same location.

mapcat:
@yakra: Correct, which is why I changed the labels back before submitting the request. But this has not been applied universally, which is what confused me at first. I wish I could remember where I saw the intersecting route name taking precedence over the state line, but can't at the moment.

yakra:
Oops, didn't see that last commit.

PR #888, opened 2016-11-03
Bay state NMPs part 2

yakra:


yakra:
One more commit to possibly add

yakra:
Ready to rock.

mapcat:
@yakra: from your experience, how many times have the files in http://tm.teresco.org/nmp_merged/ included false positives? I'm going to get started on a NMP purge in my states and am curious whether I should go through each file that comes up in nearmisspoints.log point-by-point or just let the auto-generated files replace the existing ones. So far, only checking Ohio, there are hundreds of routes with NMPs, so anything that could streamline the process would be welcome.

yakra:
I have actually not looked at the files in nmp/merged/ at all. I like to look at the data in excruciating detail and make sure everything is *just so*, so I took the more time consuming route. I used OpenOffice to filter tm-master.nmp and make smaller NMP files for each region I'm working on. So I guess I'm taking the "going thru point-by-point" route, but having the visual aid of the HDX really streamlines the process. Having a TMG file loaded into the HDX in another browser tab is also helpful.

How many times have I found a **false positive**, though? Not many.
I would define a "false positive", here, as an NMP that *doesn't look intentional*, and thus gets merged into a single point, but should not have been merged. I see two varieties of these:
• False-multiplex breakers (such as frontage roads, etc.) that should *look intentional*, but don't, and need to be adjusted to be off by 0.000001. I only found **one** of these in TX, out of what started out as 89 near-miss pairs. I don't *recall* there being any in the other states I've done yet.
• Pairs that shouldn't look intentional, and are just two very closely spaced points that should not be merged. Prince Edward Island has 10 NMP pairs, and I suspect a bunch of them may be closely-spaced junctions as strip-mined from the GIS shapefiles. I threw pe.nmp in my "procrastinate" folder for this reason. :)

But then, there's the **false negatives**...
Points that do *look intentional*, and thus won't be merged, but should be. These are bona fide errors that will still need to be fixed manually. I've seen plenty of examples of this, often probably resulting from "the mysterious 'point drift' that affected the Interstate & US Routes in the early years of CHM." Picture some interchange between I-95 & US1 in New England and you'll get the idea.
So, even if you first replace everything with the *nmp/merged/* files, another pass thru (via the HDX or whatever method you end up using) would still be in order. False negatives have to be fixed manually either way.

• The upside of getting the *nmp/merged/* files in is that the visual signal-to-noise ratio in the HDX is improved; you just see the negatives. Then, it's just a matter of going thru & fixing the false negatives, and leaving the true ones alone, probably marking them as FPs in nmpfps.log. (I have not started that last step yet.)
• The downside is that false positives, while they may be relatively rare, will be merged and thus not show up if you load the resulting NMP file into the HDX. Without having that visual aid, I think it'd be harder to point out & identify & know to fix them. You've got to have a way of keeping track of them and knowing where they are, and backing out incorrectly merged points, not copying in files that contain them, or editing up new files that keep the new merged true positives & the old original false positives, mix-n-matched as needed.

The trade-off: Is the convenience of not having to worry about manually fixing the true positives worth the risk of passing over false positives that would need to be maintained or corrected?

With NMPs being relatively sparse in my states, I chose to go the more cautious and time consuming route, just look over everything point-by-point in the HDX, and replace points manually. Again, I like to make sure all the data are *just so*. It gave me the opportunity to resync a bunch of points from the very early years that were more significantly off the centerline; sometimes just replacing a bunch of points along both parallel I-95 & US1 was, if not easier, just *nicer*.
I started with the low-hanging fruit (nb.nmp had 2 near-miss pairs for one single point; RI & NS had 5 near-miss pairs each...) and worked my way up by filesize, leveling up and becoming faster & more efficient at the process.

Your threshold for "How sparse are the NMPs in my state?", and for the amount of work potentially involved, could be anywhere. I'm getting 130 near-miss pairs for Ohio right now. I just finished Texas, my largest single-state NMP file, with 89 near-miss pairs.

--

If you do as I did and choose the more thorough route, as far as anything that could streamline the process...
• As mentioned above, I used OpenOffice to filter tm-master.nmp and make smaller NMP files for each region. Then, I insert an extra line of whitespace every third line, to better visually distinguish the point pairs when looking at the NMP file in my text editor. When I make the edits to fix a point (or find an intentional FP-NMP), I remove the entries from the NMP file and reload it into the HDX. This cuts down on the visual clutter, leaving me with what still needs attention. I find this makes my plan of attack easier to see as work progresses.
• First thing I like to find is points with 3 (or more) routes. If, say, US 5 & SR 10 have matching coords and I-91 mismatches, I'll just edit the one file to match the other two. Keeping to fewer edits in fewer files this way has been one thing that's helped keep me sane.
• Having the HDX open in two browser tabs, one with the TMG file & one with the NMP file, has been helpful for me. It helps me to pick out points where 3 or more routes come together, places where multiplexes are involved, which points do/don't match with what at a given junction, and to identify what route files I'll be loading and looking at.
• Sometimes I'll load up all 3+ routes for a given point in the HB in separate browser tabs, and click on the point label in question. It centers the map there, and displays the coords in the InfoWindow. Some quick Ctrl+Tabbing around (Or, Alt+1, Alt+2, Alt+3 etc. in Firefox on Linux, or, Alt+1, Alt+2, Alt+3 etc. in Firefox on Windows) and I can quickly see what coords do/don't match on what routes in the InfoWindow. This works better for me than just trying to pick the info out of the NMP file itself and having my eyes start to glaze over.
• Next thing I might do is just do a bulk replacement of everything in two files that are multiplexed (E.G. MA I-291/US20) or two files that parallel one another for long stretches (E.G. CT I-95 & US1 or MA I-91 & US5).
• Relatively isolated points in more rural areas (E.G. TX25 & TX240) without any other points on either route to worry about are pretty easy to pick off.
• A single route with several NMPs along its length, I might leave alone for a while as I pick off other NMPs in the region, till the path becomes more clear. This ties into how I like to make the edits to the fewest files, when I can. For example, by resyncing MA I-95 exits 4 & 6, I avoided having to edit I-295 & I-495. Other I-495 NMPs were taken care of by editing the intersecting routes, which were getting other edits anyway.
What I'm saying with the I-495 example, is that if a route has several NMPs, you may not have to edit it, just due to editing the other intersecting routes instead.
On the flipside, making several edits to a single file can help avoid having to edit several others.

jteresco:
Wow.  That needs to be in the forum instead of buried here in a GitHub pull request.  Thanks for the excellent description and for taking the time to do it all!

yakra:
Cross posted to the forum at roughly the same time.

PR #899, opened 2016-11-05
AUTB review (B100-B230)

michihdeu:


michihdeu:
@jteresco , please update the graphs because I wanna check AUTB. Thanks in advance!

jteresco:
Update being launched soon will include new graphs and nmps.

PR #916, opened 2016-11-09
Bulgaria

si404:
2016-11-09;Bulgaria;E79;bgr.e79;Removed from II/18 and relocated onto SST between II18_E and II18_N
2016-11-09;Bulgaria;E80;bgr.e80;Removed from II/18 and relocated onto SST between II18_N and A2


michihdeu:
bgr.i002.wpt: A2/I4 should be renamed to A2 only
bgr.i004.wpt: A2/I2 should be renamed to A2 only
--> you've made the modification for I4 but not for I2

bgr.ii056.wpt:
II56_S http://www.openstreetmap.org/?lat=42.633043&lon=25.193023
+II56_N http://www.openstreetmap.org/?lat=42.633343&lon=25.198281

You've made a self_ref error...

si404:
> bgr.i002.wpt: A2/I4 should be renamed to A2 only
no it shouldn't. it would be A2_suffix, and there's no rule against having labels with two different classes in them. I don't like them when you lose the prefix, but 

Can we also save that type of mundane datacheck error fix for just doing it, rather than cluttering up github. Stuff like "this won't work", yes, but stuff like "oh no this will make an error on a preview system that you will easily spot as its flagged but you must create it stat" or "you didn't do this thing I asked you to even though the status quo is perfectly fine with the manual" is best off not bothering with/putting in the forums when it's live on the site.

PR #941, opened 2016-11-22
canmb -> devel

yakra:
in preparation for canmb -> devel
details at http://tm.teresco.org/forum/index.php?topic=9.msg4356#msg4356


yakra:
I'm going to add some commits to bring in the 1st batch of canmb -> devel, and then reopen.

yakra:
The first batch of canmb routes (former cansph routes & children) is now ready to go. Reopening.

PR #943, opened 2016-11-23
reviews

si404:


si404:
2016-11-23;Greece;A7;grc.a007;Route extended at south end from exit 18 to exit 21
2016-11-23;Finland;Vt19;fin.vt019;Route removed from St624 and Kt67 and relocated onto Seinajoki bypass between exits 4 and 9
2016-11-23;Finland;Kt67;fin.kt067;Route extended at west end from St624 to Vt19

PR #950, opened 2016-11-26
Reroute, add additional waypoint (WA 92)

compdude:
Rerouted along Quarry Road (Granite Falls Bypass) with three new waypoints added; also added one at 84th St NE.

### DIFF
--- a/hwy_data/WA/usawa/wa.wa092.wpt
+++ b/hwy_data/WA/usawa/wa.wa092.wpt
@@ -1,4 +1,7 @@
 WA9 http://www.openstreetmap.org/?lat=48.027849&lon=-122.110173
 MacRd http://www.openstreetmap.org/?lat=48.033538&lon=-122.053202
-GraFalByp http://www.openstreetmap.org/?lat=48.083258&lon=-121.990181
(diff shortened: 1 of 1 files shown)


yakra:
* GraFalByp is in use, and should be included as an AltLabel for GraFalHwy.
* GraAve, also in use, could also be included as an AltLabel for GraFalHwy, so as not to break existing list files. Anyone using this point would have to have another point farther west as the other point in their list file, and would thus not have anything erroneously mapped that they have not actually traveled.

Otherwise, looks good. Once the above is sorted, updates text could read something like
2016-11-26;(USA) Washington;WA 92;wa.wa092;East end removed from Granite Falls Highway between Quary Road and Granite Avenue, and relocated onto Quarry Road between Granite Falls Highway and Mountain Loop Highway.

jteresco:
Will also hold off on this one until those additional changes are put in.

compdude:
Okay, I added in the alt labels. Hopefully I did it right...

yakra:
The AltLabels look good. Merging this one in. I'll leave the updates text, in my previous comment, for @jteresco to handle.

PR #951, opened 2016-11-26
WA 823 rerouted onto Jim Clements Way

compdude:
Source: http://www.wsdot.wa.gov/projects/sr823/selahvicinity/


yakra:
Looks good to merge.
2016-11-26;(USA) Washington;WA 823;wa.wa823;Removed from 1st Street and Naches Avenue, and relocated onto Jim Clements Way between 1st Street and Naches Avenue.

PR #949, opened 2016-11-26
Change "Seattle Freeway" to "West Seattle Bridge"

compdude:
This is the actual name for the freeway. Note that the "West" is not a directional prefix, but it refers to the West Seattle neighborhood that the freeway serves.

PR #956, opened 2016-11-26
Rename

compdude:
Was named "seafwy.wpt"


yakra:
Per the discussion at #949, this file rename looks good. I'll continue the discussion over at #949.

PR #972, opened 2016-12-01
CZE

michihdeu:


jteresco:
Looks like there's a file introduced/modified here that has at least one non-ASCII character, so I'm commenting out from systems.csv for now.  I can investigate within the next few days if it's not taken care of by the time I have a chance.

PR #975, opened 2016-12-02
CZE II647 I11/Î58 --> I11/I58; czeii devel

michihdeu:
Sorry, I made a check of all files. The last new file, II647, had a typo (I11/58) and I 've added an I just before the upload. For any reason, I've pressed the ^ key which makes a Î...


jteresco:
Thanks for the fix.  I'll pull this in later when I'm ready to run the next update.

PR #981, opened 2016-12-05
NLD A205 + A208 + E25 / Data check

michihdeu:


michihdeu:
2016-12-05;Netherlands;A205;nld.a205;New route
2016-12-05;Netherlands;E25;nld.e25;Route removed from road through
Maasdijk and onto A20 between 'N220' and '6(A20)'

PR #982, opened 2016-12-05
Review

si404:
2016-12-05;(USA) Wyoming;US87 (Cheyenne);wy.us087;Route truncated at north end from the Montana border to WY193, and then extended at north end to the washout.
2016-12-05;(USA) Wyoming;US87 (Sheridan);wy.us087she;New Route


jteresco:
So there's a permanent gap on US 87 in Wyoming?  I had no idea.

the-spui-ninja:
Yeah, it's been there for years: http://www.aaroads.com/west/us-087_wy.html

Note: the article says "sometimes signed as WY 344"; I've never seen WY 344 signed at all.

si404:
On inheriting WY summer '13, moved US87 off of I-90 and onto its actual route, I split the routes, raised the issue, and undid the split as Tim vetoed it (and not for bad reasons). However it is clear now that Tim's notion that WY were beginning to move US87 onto WY193 (without telling AASHTO, who denied such a request in the 90s) have not come into fruition.

jteresco:
Looks like it.  I'll fix.

PR #991, opened 2016-12-10
KY updates

mapcat:
2016-12-09;(USA) Kentucky;Bert T. Combs Mountain Parkway;ky.commtnpkwy;Route moved to new alignment between exits 72 & 75
2016-12-09;(USA) Kentucky;I-66Fut (Bowling Green, KY);ky.i066futbow;Route truncated to I-65 at western end
2016-12-09;(USA) Kentucky;KY 3050;ky.ky3050;Route added


mapcat:
Thanks Jim. Thought I'd missed tonight's update by less than 5 min!

jteresco:
You did, but I saw it come in and restarted.  Have to restart again - there was an error.  Hope I resolved it.

yakra:
> 2016-12-09;(USA) Kentucky;Bert T. Combs Mountain Parkway;ky.commtnpkwy;Route moved to new alignment between exits 72 & 75

Inclusion of the "pivot waypoints" (72 & 75) is good. Best practices, as handed down from the CHM days, was to include a short description of what the route was removed from and what it was relocated onto. Absent any other stable road names here, something like "closed roadway partially paralleling KY 30 at its west end" and "straightened, direct connection" may do.
Though I know from experience that, sticking to the strictest interpretation of Tim's CHM-era rules can quickly make for some wordy updates entries, that can at times be a real PitA to write. As such, it's not earth-shattering if your choose not to do this. I do see some value in it, though; say, if someone's reading the text and trying to find out how an alignment changed years down the line...
Lord knows, for more recent updates after moving to TM, I've deviated from these rules myself plenty. Such as a omitting the "pivot waypoints" when a route is removed from a new business route; in many cases this can be sufficiently obvious IMO.

mapcat:
I considered writing more, but couldn't come up with anything meaningful. It's not clear what's happening with the abandoned alignment (a curve that's been straightened). Would changing the entry to "moved from curving former alignment to straighter alignment to the south" be more helpful to users in determining whether to de-clinch the old segment? When reviewing updates related to my own travels, I always check maps when making this determination, since something visual is usually a lot more useful* than even the most accurately worded statement.

*somewhere in the order of  10^3

jteresco:
I for one am happy we have relaxed the updates entry rules a bit but we do want the information in there.

The @mapcat comment about the visual being so much more useful makes me think we really want to have some way to show old and new alignments in map form.  For some updates, I've found myself at a loss to figure out exactly what I need to do to correct my list file for realignments.  I am not proposing a good mechanism for this, but it's something to consider.  First thought: before and after screen grabs we can link from the updates entry in cases where this seems useful?

mapcat:
@jteresco I'd love that...seems like a good idea to work out. How would you suggest storing them/making them available?

jteresco:
Let's move this discussion over to the forum.

http://tm.teresco.org/forum/index.php?topic=1849.0

yakra:
This was one of those weird scenarios where I was arguing a point I'm not even necessarily in favor of... I also find it helpful to look at the HB to get a visual on what's happened with changes like these. Once I looked at the HB in this case, it was immediately apparent what was going on.

PR #1007, opened 2016-12-15
TN usansf extensions

mapcat:
2016-12-15;(USA) Tennessee;TN 111;tn.tn111;Route renamed and extended to US127 near Kentucky border
2016-12-15;(USA) Tennessee;TN 155;tn.tn155;Route extended to US70S on west end and to US31 on east end


yakra:
**Extend TN155:**
• The road between US31 & US70S is missing. Is that not actually part of TN155?
• The "US31A/41A" label: using just the 'A' suffix has never been preferred for bannered US routes, even in jurisdictions where it's signed as such (such as my own Maine) -- the standard has been to use the full banner, E.G. US201Alt.
For my own personal opinion here, I've never liked including two bannered routes in a label (E.G. US31Alt/41Alt -- just looks a bit clunky IMO), though AFAIK there's no hard & fast rule against it. (What bothers me *more* though is having one mainline & one bannered US route in the same label, E.G. US2/201Alt; those can be ambiguous & confusing IMO.)
That said though, if you do choose to go with "US31Alt/41Alt" it won't bother me all that much. :)

A statewide thought: If USATN is to be all one system, for purposes of consistency, what to do about TN**s**155 style labels?

mapcat:
TN155 signage ends at US31 & US70S.
I'll fix 31A/41A.
Re TNsX labels: I've been waiting for James to get involved in the forum conversation and planned to ask him there.

PR #1020, opened 2016-12-21
KY/TN fixes + usatn 1-50

mapcat:
OK if I wait to create the .csv files until after usatn is done? I may tweak names, combine/split routes, etc based on further general comments in the forum.


jteresco:
You are welcome to add wpt files to the repository without csv entries, but they won't be ingested into the DB unless they're mentioned in a csv, meaning you won't see them in TM's highway browser.

mapcat:
That's fine. Thanks.

PR #1009, opened 2016-12-17
Extend WA 103 northward, adding WA 100 Spur

compdude:
WA 103 is being extended northward per this forum post: http://tm.teresco.org/forum/index.php?topic=1791.0

Also, I am adding WA 100 Spur Fort Canby per this forum post: http://tm.teresco.org/forum/index.php?topic=1790.0

PR #1039, opened 2017-01-04
Extend WA 103 northward, add WA 100 Spur

compdude:
This is a repeat of pull request #1009, but using an updated master fork.


jteresco:
Looks like the wpt names and the csv entries don't match, causing the site update to fail:

`FileNotFoundError: [Errno 2] No such file or directory: '../../../HighwayData/hwy_data/WA/usawa/wa.wa100sprfor.wpt'`

Unfortunately, I can't easily investigate or run another site update now.  Please fix as needed and submit a new pull request that I can bring in over the weekend.

compdude:
Oops, I forgot to create the wpt file for WA 100! That's most likely the problem.

PR #1072, opened 2017-01-21
Oregon revisions: First set

Bickendan:


jteresco:
@Bickendan Looks like you found the way to submit a pull request.

Since this is @Bickendan's first time in GitHub, could anyone else give a quick look over before I pull in?  Missed tonight's site update anyway, so any time Saturday would be good.

ovoss:
I won't have time for more right now, but I didn't see any Updates entries in the pull request, such as the one needed for the rerouting of US 20 in the Eddyville area (which will affect my list file).  We usually include items for the Updates list as comments in the pull request, though Bickendan may've been planning on giving them to Jim some other way.


The part of the US 20 route file I looked at, between US 101 and I-5 (to figure out how my list file would need to be changed) seems to have many more shaping points than necessary.


On January 20, 2017 at 10:32 PM Jim Teresco <notifications@github.com> wrote:


@Bickendan https://github.com/Bickendan Looks like you found the way to submit a pull request.

Since this is @Bickendan https://github.com/Bickendan 's first time in GitHub, could anyone else give a quick look over before I pull in? Missed tonight's site update anyway, so any time Saturday would be good.

jteresco:
Since the reordering of updates.csv by region, updates entries can now be included right in pull requests with very little risk of introducing conflicts.

Bickendan:
Anything else needed from me then before this is all updated? @ovoss The US 20 update does include the reroute off of now Crystal Creek Loop.

yakra:
The DIFFs look wicked weird -- everything is highlighted red/green as a deletion/addition, rather than just isolating the parts of the file that got changed. Makes it hard to see the changes. Not @Bickendan's fault, mind you, probably a software thing. Just thought I'd mention.

As @ovoss noted, I'm also seeing LOTS of extraneous shaping points. For example, I was able to whittle OR6 down from 70 to 29 waypoints and keep it within tolerance.

yakra:
Hm. The DIFF for or.or039.wpt is funny-looking -- I see a bunch of ^M characters at the end of each line.

michihdeu:
Minimum or.or018buswil.wpt, or.or082trkelg.wpt, or.or086s.wpt, or.or204trkelg.wpt and or.or414.wpt contain illegal characters.
I think it's CR which must be removed (I use Notepad++, open all files with one click, select "show hidden characters", find+replace \r by nothing in all files with one click).

Bickendan:
Very odd on hidden characters -- I do all my edits via the .wpt editor and copy/paste into Notepad.

michihdeu:
I do it the same way. However, my wpt files always contain CRLF after sync from GitHub. I always have to replace CR.

jteresco:
It's so annoying that after all these years, we're still fighting the CR vs. CR/LF problem...  I'd like to see if we can resolve that part before merging.

As for the excessive shaping points, those aren't introduced by these changes, right?  So the routes could be thinned out later.

Bickendan:
What do I need to do on my end for the CR/CRLF problem? 
And shapepoints are mostly as they were before revisions. I'll do a shaping pruning pass after the rest of the set has been tweaked.

michihdeu:
> What do I need to do on my end for the CR/CRLF problem?

I read in GitHub help that there might be an option but it didn't work. I gave up and do it manually by replacing it with Notepad++.

jteresco:
I'm not likely to be too helpful, since I don't work in Windows.  On my Mac or on FreeBSD, when someone submits updated files by email that have LF problems, I have a Python program I run to strip them out.

michihdeu:
It takes less than 5s with Notepad++ for - let's say 5,000 files - if they are plain in the same folder. Including opening and saving. No need for implementing a program...

yakra:
> shapepoints are mostly as they were before revisions

The diffs say otherwise, but yes, I agree with a shaping pruning pass.

jteresco:
I'm thinking I'd like to merge this in and just live with whatever the file format issue is.

ovoss:
Typo fix needed for US 20:  CreCreLp_E => CryCreLp_E (matching the corresponding point on the other end of the loop).

yakra:
https://github.com/TravelMapping/HighwayData/pull/1072#issuecomment-275211912
> On my Mac or on FreeBSD, when someone submits updated files by email that have LF problems, I have a Python program I run to strip them out.

Is this a specific standalone program, or has this functionality been integrated into siteupdate.py?

Was all of this only an issue before switching to using localupdate.sh on noreaster?

jteresco:
> [#1072 (comment)](https://github.com/TravelMapping/HighwayData/pull/1072#issuecomment-275211912)
> 
> > On my Mac or on FreeBSD, when someone submits updated files by email that have LF problems, I have a Python program I run to strip them out.
> 
> Is this a specific standalone program, or has this functionality been integrated into siteupdate.py?
> 
> Was all of this only an issue before switching to using localupdate.sh on noreaster?

I have a program called crlf.py (Python 2):

`#! /usr/bin/env python

"Replace CRLF with LF in argument files.  Print names of changed files."

import sys, re, os
for file in sys.argv[1:]:
    if os.path.isdir(file):
        print file, "Directory!"
        continue
    data = open(file, "rb").read()
    if '\0' in data:
        print file, "Binary!"
        continue
    newdata = re.sub("\r\n", "\n", data)
    if newdata != data:
        print file
        f = open(file, "wb")
        f.write(newdata)
        f.close()
`

I don't remember where I got it, but I've had it for years, likely predating TM.

When one of our email-based highway data contributors sends me an update, I put the files in a clean directory, and start by running the above program to ensure Unix-style lines.  Next, chmod 644 to make sure permissions are good.  Some files extract with all kinds of crazy permissions.  Then I do a diff on each one with the existing version of the file to make sure there aren't more changes than indicated.  This also helps when someone mistakenly renames a file or something like that.  Once each diff passes a quick visual inspection, I move the file into place.  When done, I commit and push the changes.

The main reason for fixing the line feeds is to prevent Git from thinking the whole file had changed, when it would only be line feeds differing on almost all lines.

I don't believe siteupdate.py will care.  It's Git that might care.

yakra:
> The main reason for fixing the line feeds is to prevent Git from thinking the whole file had changed, when it would only be line feeds differing on almost all lines.
>
> I don't believe siteupdate.py will care. It's Git that might care.

Yup. As we saw with the DIFFs in this pull request.
So, these may still squeak in occasionally, for people using Git(Hub) itself rather than emailing. If it does, I suppose, not the end of the world. While the DIFFs may be less than human readable at one particular point in the commit history, life should go on afterwards as a contributor goes on to make edits using the same OS. On subsequent commits, we should just see the individual, localized changes in the DIFFs.

PR #1074, opened 2017-01-22
BEL/NLD

michihdeu:


jteresco:
We have two entries in systems.csv for belr, one active and one devel.  Which one is correct?

jteresco:
BTW, can't run a site update with both since it loads the csv twice and that causes all kinds of trouble.

jteresco:
I now see there's a beln directory - assuming that's what the devel entry should be.  Let me know if that's incorrect.  That's what I'll do for the site update a little later unless I hear otherwise.

michihdeu:
Sorry and thanks for the fixes.

I think an automated check should be a priority for 2017/18: https://github.com/TravelMapping/DataProcessing/issues/22.

PR #1078, opened 2017-01-22
reviews and system starts

si404:


jteresco:
Commenting out estt for now to get the update through:

estt.................................Traceback (most recent call last):
  File "./siteupdate.py", line 1813, in <module>
    r.read_wpt(all_waypoints,datacheckerrors,args.highwaydatapath+"/hwy_data")
  File "./siteupdate.py", line 611, in read_wpt
    with open(path+"/"+self.region + "/" + self.system.systemname+"/"+self.root+".wpt", "rt",encoding='utf-8') as file:
FileNotFoundError: [Errno 2] No such file or directory: '../../../HighwayData/hwy_data/EST/estt/est.t048.wpt'

Not sure if the "(not done)" notation means it was not supposed to be in the submitted csv.

jteresco:
The file srb.e70.wpt had two consecutive lines with the same coordinates.  I removed the new one to get a site update past that point.

PR #1084, opened 2017-01-24
Split WA 99 into two sections

compdude:
WA 99 is being split into two sections to reflect the fact that it has a gap in Tukwila. The northern section is going to be called wa.wa099sea.wpt, and the southern section will remain as wa.wa099.wpt. See forum thread: http://tm.teresco.org/forum/index.php?topic=445.0


jteresco:
This one definitely needs to include a good updates entry.

compdude:
Oh yeah, forgot about that...

compdude:
Okay, how does that update entry look to you?

jteresco:
I'll let others make a final call, but it seems reasonable.

jteresco:
I'll merge this in for tonight's site update, but we can still update the updates entry if deemed necessary.

yakra:
The northern (Seattle) should be the main WA99 segment (no Abbreviation), as it's the longer segment.
The southern segment should be the one with a City/Abbrev. Des Moines, perhaps?

jteresco:
I'm making the change suggested by @yakra, but using Federal Way as the city/abbrev.

PR #1097, opened 2017-02-01
late Jan changes

si404:


yakra:
"mister peanutbutter"?

si404:
It is a Bojack reference. He's a Labrador, from the Labrador peninsular, so what better name for the commit that includes first drafts of the NL5xx series? Other than the descriptive boring one of course ;).

PR #1101, opened 2017-02-03
cannl -> preview

si404:


yakra:
Did you end up making any changes to TCH 1?

si404:
No changes to TCH Main or NL1 as of yet, IIRC. Any changes to NL1 would be
visible in the browser (NMP, etc) now the system is preview, if you want to
check. If I make changes, I'll let you know and you can sync the files up.


Si

PR #1114, opened 2017-02-08
NY tweaks

yakra:


yakra:
This should probably be merged before #1115.
If #1115 is merged first, then this can just be closed

PR #1123, opened 2017-02-15
Ireland rehash done

si404:
Other than the load of datacheck and NMPs I've added!


jteresco:
Error:

FileNotFoundError: [Errno 2] No such file or directory: '../../../HighwayData/hwy_data/SCT/gbrtr/sct.dorembocr.wpt'

PR #1134, opened 2017-02-26
ESP

michihdeu:


michihdeu:
n001bui.wpt begins with a hidden point (because I wanna ask Si if the exit location should be changed). Not sure if this might cause a problem. If it does, please remove the hidden point. Thanks!

PR #1177, opened 2017-04-01
Remove 176thSt waypoint from WA 509

compdude:
Exit has been closed for several years.


rickmastfan67:
Wouldn't marking it as 'closed' have been a better option since it was a former exit instead of deleting it?

compdude:
What do you mean by marking it as closed? I thought it was OK to delete it since nobody was using it according to the pointsinuse.log file.

rickmastfan67:
Closed/former exits are marked with a * in front of them.  Especially if the route was already active on the site.  Take a look @ I-99 in PA as an example where the route 'temporarily' ended for awhile in Skytop as US-220/US-322. http://tm.teresco.org/hb/?r=pa.i099

compdude:
I still don't understand the need to keep closed exits in the HB when the point was not in use by anyone. In this case, if the exit re-opens at a later date, it's not that difficult for me to add in the point. And regardless, I don't think it will open again. According to the historical imagery on Google Earth, it was built during the construction of SeaTac Airport's third runway, presumably for construction access, and was closed soon after the third runway opened.

compdude:
After thinking about it some more, I can see the logic of keeping this as a "closed" exit. We could have a new user come along who might have driven this road when the exit was open and gotten off there. So, in order to allow him to accurately document his travels, we should keep this in there as a closed exit. I'll do another pull request soon and re-add this waypoint as a closed waypoint.

PR #1181, opened 2017-04-02
Updates in Idaho

compdude:
Realigned US 95 in Athol, also added some new exits south of there, (http://tm.teresco.org/forum/index.php?topic=1926.msg5607#new) and added US 95 bypass of Sandpoint, ID. Also moving waypoint at US 95/ID 58, and updating position of a waypoint on US 2 in Dover.

EDIT: Added exit 328 to US 20

### DIFF
--- a/hwy_data/ID/usaid/id.id054.wpt
+++ b/hwy_data/ID/usaid/id.id054.wpt
@@ -1,7 +1,7 @@
 ID41 http://www.openstreetmap.org/?lat=47.953093&lon=-116.868114
 RamRd http://www.openstreetmap.org/?lat=47.948047&lon=-116.807268
 OldHwy95 http://www.openstreetmap.org/?lat=47.947800&lon=-116.708225
-US95 http://www.openstreetmap.org/?lat=47.947842&lon=-116.699009
+US95 http://www.openstreetmap.org/?lat=47.947806&lon=-116.696209
 GoodHopeRd http://www.openstreetmap.org/?lat=47.947430&lon=-116.612341
 BlaCirDr http://www.openstreetmap.org/?lat=47.970497&lon=-116.566934
 3rdSt http://www.openstreetmap.org/?lat=47.979401&lon=-116.563801
--- a/hwy_data/ID/usaid/id.id058.wpt
+++ b/hwy_data/ID/usaid/id.id058.wpt
@@ -1,4 +1,4 @@
 WA/ID http://www.openstreetmap.org/?lat=47.434908&lon=-117.039947
 DreRd http://www.openstreetmap.org/?lat=47.434356&lon=-117.032821
 +X01 http://www.openstreetmap.org/?lat=47.425478&lon=-117.012441
-US95 http://www.openstreetmap.org/?lat=47.427688&lon=-116.985834
+US95 http://www.openstreetmap.org/?lat=47.423950&lon=-116.986520
--- a/hwy_data/ID/usaid/id.id200.wpt
+++ b/hwy_data/ID/usaid/id.id200.wpt
@@ -1,4 +1,4 @@
-US95 http://www.openstreetmap.org/?lat=48.290004&lon=-116.550661
+US95 http://www.openstreetmap.org/?lat=48.290318&lon=-116.548977
 KooCORd http://www.openstreetmap.org/?lat=48.309106&lon=-116.529925
 MainSt http://www.openstreetmap.org/?lat=48.309103&lon=-116.513636
 ShiMillRd http://www.openstreetmap.org/?lat=48.316013&lon=-116.488847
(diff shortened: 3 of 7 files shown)


ovoss:
The realignment in Athol doesn't have. as I think it should, waypoints for the closed intersections where the new alignment peels away from the old one. I think the realign is borderline significant enough for an entry in the Updates table, indicating where the new alignment begins and ends (which is where waypoints at those locations), so users can decide whether and how to adjust their list files.

The realign in Sandpoint appears to have such points already. However, shaping point x47 is unneeded and can be deleted. I haven't yet checked the other realigns on US 95.

jteresco:
I'll wait to pull this one in until Oscar's concerns are addressed.

compdude:
Looks like I'll have to do the same thing with the realignment near the ID 58 junction as well.

compdude:
Okay, now it _should_ be all A-OK now.

ovoss:
A few cleanup items on the US 95 changes.

1. There's no Updates entry for the US 95 realignment near ID 58.

2. Nor is there a point for where northern end of the old alignment near ID 
58 (there is one for the south end). Satellite view indicates that the north 
end is just south of the intersection with W. Missile Base Rd. (that name is 
shown on Mapnik if you zoom in far enough). I suggest a new MisBaseRd 
waypoint there, and that you use it in the new Updates entry.

3. US2/ID 200 => US2/200 -- even if the second number is for a route in a 
different system, just use the number.

4. The Updates entry for the US 95 realignment in Sandpoint indicates the 
new alignment splits from the old alignment at the (now closed) intersection 
with Cedar St., but there's no waypoint there. I had assumed. from my 2012 
offline mapping software (Microsloth Streets and Trips, now discontinued), 
that the peel-away point was at Superrior St., for which there is a point, 
and that the bypassed part of US 95 included Superior St., 1st St., and a 
couplet of one-way streets Cedar St. and Pine St. I suggest revising either 
the Updates entry, or the route file, depending on what more accurately 
reflects the old alignment.  I plan on using the SupSt point in today's list 
file update, so please keep that whatever you do.

To my constenation, I visited Sandpoint twice last summer (leaving and 
returning via US 2, rather than through Coeur d'Alene), and managed to miss 
the new bypass both times.

----- Original Message -----

compdude:
For the realignment near ID 58, it actually started near where I placed a new waypoint, at EldRd. The two points south of there were already there; it was only the junction with ID 58 that was in the wrong place. And BTW, this realignment and addition of that folded diamond interchange with ID 58 took place in 2009, according to Google Earth's historical imagery, so if you drove the road since then, you can still consider this section clinched. :) 

I'm still not 100% sure if I need an updates entry. When I moved a couple waypoints on WA 522 to reflect a couple realignments in pull request #948, no one told me I needed an updates entry. Here, I'm doing pretty much the same thing.

As for (3) and (4), I'll fix those things.

ovoss:
Thanks for the clearer Updates entry for US 95 in Sandpoint.

I think I might be confusing several US 95 realignments south of I-90 in the 
general area of ID 58, some of which might have happened a long time ago and 
were reflected in the original US 95 route file when Clinched Highway 
Mapping expanded to cover US routes in the late 2000s.

There's the ID 58 waypoint tweak, which is both very recent and not 
Update-worthy. Then there's the one starting between there and W. Missile 
Base Rd., and ending at Old US95_S in Mozart. Then there's a third from just 
south of EldRd to just south of SetRd. The last two seem to predate TM and 
maybe also CHM (though OldUS95_S came after CHM, which might've thrown me 
off on how recent was that realignment). We ordinarily don't create 
waypoints for such old realignments unless we need such points for shaping 
or avoiding distance errors, and definitely do not include them in Updates.

As for whether more recent realignments are Updates-worthy, such as what you 
mention with WA 522 (which I'm not familiar with), that's a judgment call 
about what users would care about enough to adjust their list files, and 
there's no bright lines there. FWIW, when you first brought up the Athol 
bypass, I called that "borderline" for an Updates entry, and I'm one of the 
fussier people about such things. When I-80 in California was moved from the 
old to the new Bay Bridge between San Francisco and Oakland, I put in 
closed-intersection waypoints at each end where the new alignment peeled 
away from the old one, taking advantage of now-defunct open-source satellite 
imagery. Nobody is using those points, apparently treating the new bridge 
(as I did) like just another non-declinching bridge replacement. This might 
be worth some more discussion in the forum.

As for my travels on US 95 in Idaho, I have no idea when I traveled the 
segment of US 95 between US 195 and I-90, though most likely before 2008, so 
to be conservative I've treated everything between OldUS95_S and SetRd (and 
in the next update, up to EldRd) as unclinched, especially since I might go 
back to northern Idaho in the not-too-distant future. I've been doing road 
trips for more than three decades, and my records of my pre-2007 travels are 
spotty and largely just a big wall map in my kitchen. I know I traveled US 
95 from I-90 to north of Sandpoint in 2008 (just enough to snag Idaho's 
northernmost county) and from US 2 in Sandpoint to the Canadian border in 
2016.

----- Original Message -----

PR #1211, opened 2017-04-21
datacheck

si404:


jteresco:
Lots of typos here, working through them.  Most obvious, but I want to point out that I'm deleting prt.ic002 from the csvs since there's no wpt.

PR #1214, opened 2017-04-24
I-22

si404:
Makes I-22 a proper interstate, per forums. Check with Froggie before pulling as these are his states.


jteresco:
I'll give this a few days for discussion.  Please comment in the forum when you have a chance if you have not already done so, @ajfroggie.

PR #1228, opened 2017-05-01
portugal fixes

si404:


michihdeu:
I think you forgot removing IC4 from csv files

PR #1232, opened 2017-05-03
canbc finished?

si404:


ovoss:
I haven't finished reviewing mapcat's suggested changes to TCHMai and its counterpart BC1. I suggest the pull request go through as is, but I'll complete my review using the files as Si has revised them. That won't happen until Friday at the earliest, as Si's revisions won't be live until tomorrow, and I'll be medically out of pocket most or all of tomorrow. If I think further revisions are needed, I'll then make a pull request for those changes. Otherwise, I'll post in the forum that my review is complete.

Activation of canbc should wait until that all is done.

si404:
absolutely oscar - my aim was to never activate without your say-so.

PR #1247, opened 2017-05-13
more canqc

mapcat:


jteresco:
I'm removing the comments in some of these .wpt files, since they cause the site update to fail.

mapcat:
Oh, sorry. I thought anything with # would be ignored. I'll keep track of the info some other way then. Thanks.

michihdeu:
You could create a Readme.md file in the same directory with comments to the system. That's the way I do it.

PR #1260, opened 2017-05-17
AZ and UT updates

si404:


compdude:
The waypoint file for Sky Harbor Blvd doesn't seem like the right one. I think you copy-pasted the wrong file into it.

PR #1271, opened 2017-05-23
Quick fix to usapr.csv and usapr_con.csv

compdude:
PR 30 was missing from both files


jteresco:
Looks like this was intentionally left out.  Closing.

PR #1275, opened 2017-05-25
Eliminate QC conflicts

jteresco:


jteresco:
@ovoss @mapcat I resolved a conflict on qc.a019.wpt, which it appears you both edited simultaneously, by picking one.  I am not sure whose it was, but I'll leave it to the two of you to decide what the route should look like.  Everything else @mapcat changed merged in clean.

ovoss:
Jim:


I sent a PM to mapcat about this. I'm OK with the point relabels he did, IF they were made to the A-19 file version I had just recently pulled into the master.  Did you stick with my version, or go with mapcat's (I can't tell which you did until you do another site update).  My version includes point synchs between A-19 and both A-40 and A-440.


Oscar

jteresco:
It wasn't obvious which was which when I resolved the conflict.  At worst, one or both of you might need to coordinate on getting a little patching up done on that file.  Thanks.

PR #1279, opened 2017-05-27
DEU

michihdeu:


si404:
Shouldn't the Hesse Landesstraßen be .wpt files, not .txt?

jteresco:
Looks like just the one type.

PR #1292, opened 2017-05-30
Added T.F. Green Connector

jjbers:


jteresco:
Did we agree to add this?  I didn't see any discussion on the forum.  If we did, there are a few problems:

1) The file needs a wpt extension, not wpk.
2) It needs to be in the subdirectory for the system in which it would be housed.
3) It needs entries in that system's csv files.

I don't want to discourage you from contributing to our highway data, but I think it's best to learn a lot more about it and do some review of our preview systems before trying to add new data.

Closing this pull request.

mapcat:
@yakra anything worth pursuing here?

Issue #1321, opened 2017-06-07
Standardize headers in data tables


Thing342:
It would be nice if all of the various tables currently in use had standardized headers, in order to aid parsing and data analysis.

michihdeu:
I think the issue can be closed now.

PR #1332, opened 2017-06-09
Standardize header for system files (#1321)

Thing342:
sets all system and _con files to use the same set of headers:

System: `System;Region;Route;Banner;Abbrev;City;Root;AltRouteNames`
Conn: `System;Route;Banner;Name;Roots`


jteresco:
Would there be any advantage to having these match the names of the columns in the DB tables?

compdude:
What does the "Root/Roots" column mean?

yakra:
root is the identifier for a single route file, the bit that goes in the *r=* field of a HB url, the bit that comes before .wpt in a filename, E.G. *me.us001*, or *me.us001altpor*.

compdude:
Oh, okay. I think what confused me was that that column was previously called "Filename" in the usawa.csv and usawa_con.csv files.

yakra:
Aa, sou. However it happened, the original creator of those files used a different variation on the header rows. Different terms to refer to different things. Which is fine -- the only purpose of the header row was for human-readable convenience, when editing the CSVs as a spreadsheet; it's not meant to be read or used by any program. I've seen some other variations in header row usage over the years. Or maybe just that one.

The syntax , as posted in the original *comment*, is the original syntax inherited from CHM. As for this pull request, the commits are all over the place, and I'm scratching my head as to its intent, direction, usefulness. That said, I rather dislike the idea of my files being edited out from under me. usame.csv, usanh.csv, cannb.csv, etc. are *mine*, and I'd prefer to just see myself in the commit history for those. :)

jteresco:
The point of the edits in this pull request, which I was planning to merge in before the next site update, is to standardize all of those, and make them match the names of the DB tables in which this information is stored.  The idea is to be able to use python libraries that will read those in more efficiently.  If there's a strong objection, I can wait for further discussion.  Since to date, those header lines are simply skipped over by the site update code, I didn't think there would be any harm in standardizing in one big batch.

yakra:
will have a specific datum in the header row provide any real advantage? Will it change the CSVs behavior WRT these Python libs at all?

jteresco:
I'll defer to @Thing342 on that, since I have not used the faster Python libraries he's proposing.  But even ignoring that, I'd say a consistency among all csv files and the DB tables is a Good Thing.  Seems to me that standardizing headers has no downside (other than having to do it, which he's already done for us) and potential advantages.

PR #1373, opened 2017-06-15
CSV and route label changes due to turnback of MN 66 & MN 228

ajfroggie:
Removed two turned back routes:  MN 66 and MN 228


jteresco:
Needs updates.csv entries?

PR #1387, opened 2017-06-18
third time lucky

si404:


jteresco:
I started removing a whole bunch of csv lines when missing wpt files were reported but I am not sure if this is the most productive way forward.

jteresco:
Well, I finished that process (removing csv entries for wpts not found) until I got a site update.  I'm going to commit and push those changes.  You can see what I removed from the commits, I hope.  In any case, we have the TravelMappingTest DB populated with your files based on railsystems.csv.

PR #1416, opened 2017-06-28
Tweaks and corrections

Bickendan:
I-5: Shifting +x59 south to just east of I-405(1A) and 299B west to match new I-405(1)
I-405: Created 1 at new location for I-5(299B) to allow accurate tracking for users who have not clinched the southernmost portion of I-405 or I-5 south of the Marquam Bridge. 
Exit sequence: 1A (northbound) Harbor Dr
1 (southbound) I-5 North/Marquam Bridge (unassigned in field!)
1B (northbound) 4th Ave
1C US 26/6th Ave
1D US 26/12th Ave
US 20: Corrected CreCreLp_E to CryCreLp_E
US 30: Added missing shapepoint (+x15a) to match I-84's shapepoint between +29 (shapepoint exit for Benson State Park -- no westbound access) and 31 (Multnomah Falls).


jteresco:
I am guessing the excessive conflicts and diffs on these files are mostly line feed problems.  How bad would it be to have to re-download the versions currently in GitHub and make your changes to those files?

Bickendan:
I thought I had done that to make the sync to US 30. But I'll do so again
just to be safe.

Bickendan:
I've downloaded I-84 into my local copy; need I do anything else at the
moment?

michihdeu:
I don't see the typical CR issue in the files (red things at the end of the line...
Anyway, in case of CR issue, re-downloading doesn't help because ALL files I download are ALWAYS with CRLF! I always have to remove the CR at the end of each line (Notepad++ has a convert feature, I just have to press CTRL+U once a file was modified just before I press CTRL+S)

compdude:
Instead of just downloading it, why not edit it straight in GitHub? That's what I do...

PR #1430, opened 2017-07-08
poland fixes

si404:


michihdeu:
@si404 : Are you sure that no update entries are required for the prta modifications? For instance, A16 has points in use...

si404:
Err, yes, because I have not yet made the change of removing part of the route, saving them for prtic activation (where there's renaming changes as well as a couple of truncations), but I made files in PRT rather than PRT/prta that can then overwrite, so that I remember and so that it won't take me long to make the changes when the time comes.

michihdeu:
ok

PR #1465, opened 2017-07-25
NMPs in Tim's old states

yakra:


yakra:
Will add more commits and reopen.

PR #1474, opened 2017-07-27
Operation NMP 500: NY

yakra:


jteresco:
@yakra, more coming?  I am about to merge these in and start a site update, but could wait a little longer if you have more in the pipeline for tonight.  Thanks.

yakra:
#1475 should be the last one I have for a while.

PR #1475, opened 2017-07-27
Operation NMP 500: KS

yakra:


yakra:
That was quick! ;)

jteresco:
Saw it come in just after starting the site update, so I killed it, merged, and restarted.  Thanks for all this NMP work, much appreciated!

PR #1502, opened 2017-08-01
DEU

michihdeu:


jteresco:
The update of the TravelMappingTest DB worked based on railsystems.  Graph generation failed, but I'm on my way out for the day and evening soon and don't have time to investigate.

PR #1506, opened 2017-08-02
US 30 concurrency fixes

Bickendan:
Synchronized I-5 to US 30 through Rose Quarter (added +x1(US30))
Synchronized I-84 to US 30, including removal of about five Columbia River Gorge exits via shapeouts (no return access to highway in opposing direction), eg, 72 -> +72
Renumbered the entire shape point sequence on I-84 to match US 30's concurrency sequence, leading to the removal of many shape points. All concurrency shape points changed to +xy(concurrent route) format; +x1(US30) on I-84 is +x1(I-84) on US 30; +x1(US30/US395) on I-84 is +x1(I-84/US395) on US 30 is +x1(I-84/US30) on US 395. 
+x1, +x2, etc, are non-concurrent shape points. 
Recentered I-84's non-concurrent exits or synchronized them to already 'cleaned up' routes; eg, OR 7, 11, 82, 86.


jteresco:
diffs look good this time, so hopefully the git merge issues are behind us.

PR #1516, opened 2017-08-05
OpNMP PA part 5

yakra:
This should almost, but not quite, complete Operation NMP PA. There should just be a few FPs to flag after this.


yakra:
parts 3 & 4 in #1510

PR #1520, opened 2017-08-07
DEU

michihdeu:


michihdeu:
No need to update the test database today. I'm almost done with deubc split. Only 182 (short) routes left. I guess most of them do not cross any border and don't have to be split.

jteresco:
OK, just let me know when you want an update of that.  The DB is small and no graphs, so the whole process just takes a minute or two.

PR #1525, opened 2017-08-09
MN 36/WI 64 Stillwater Bridge

ajfroggie:
Changes needed due to the completion of the new MN 36/WI 64 Stillwater Bridge.

### DIFF
--- a/hwy_data/MN/usamn/mn.mn036.wpt
+++ b/hwy_data/MN/usamn/mn.mn036.wpt
@@ -21,6 +21,5 @@ KeaAve http://www.openstreetmap.org/?lat=45.035625&lon=-92.903910
 CSAH5/15 +MN5 http://www.openstreetmap.org/?lat=45.036079&lon=-92.846446
 OsgAve http://www.openstreetmap.org/?lat=45.035807&lon=-92.807565
 BeaRd http://www.openstreetmap.org/?lat=45.036777&lon=-92.797136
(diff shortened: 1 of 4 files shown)


yakra:
http://cmap.m-plex.com/tools/manual_wayptlabels.php
> For numbered (alphanumeric) designations like US63, use the former designation type, like US in US 63, in a label like OldUS63, if that designation type is known. If it is not known, use a generic word like Route (Rt) or Highway (Hwy) in English or an analogous word in another language.

CHM forum discussion for context:
http://clinched.s2.bizhat.com/viewtopic.php?t=1712

I believe following the labeling guidelines & maintaining a consistency in approach (site-wide, but especially within a given route file) is something to strive for.

Thus Old64Hou should be OldWI64_Hou. The underscore, because this is a location suffix to differentiate from other OldWI64 points (IE OldWI64_W & OldWI64_E).

In the case of Old35Hou, there are no other OldWI35 points, so I'd consider OldWI35 optimal. (I can see the point of keeping a suffix on it in case other realignments make more OldWI35 points crop up in the future. If that happens though, one can always rename OldWI35 -> OldWI35_Hou then. But hey, if you wanna call it OldWI35_Hou now, I won;'t object too loudly...)

yakra:
14 hours, 12 hours... and I thought I was commenting on something that was still unmerged... o_O

PR #1526, opened 2017-08-09
Update mn.mn036.wpt

ajfroggie:
Deleted a point missed in yesterday's Stillwater Bridge update.


jteresco:
I'll get this in tonight.  Does this new bridge require an updates.csv entry?

ajfroggie:
I didn't think so.  While it's a bit of a realignment especially for WI 64, it otherwise simply replaces the previous river crossing about a mile downriver.

PR #1528, opened 2017-08-09
DEU

michihdeu:
For test DB only. I'll also submit my user list file, please update the test DB.


jteresco:
This is in the test DB.

michihdeu:
Quick, thanks!

PR #1557, opened 2017-08-24
move usansf files to usatn

mapcat:


jteresco:
Thanks - I should have noticed it was such a simple fix.

PR #1583, opened 2017-09-05
Batch pull request for NV edits

NickCPDX:


jteresco:
The main thing we need here are some good updates.csv entries for any changes deemed newsworthy.  I'm tied up a lot this week, but maybe others like @si404, @yakra, @michihdeu or @rickmastfan67 can help you get those updates entries added in an appropriate level of detail if it's not something you've had to do before.

Thanks for all the updates!

michihdeu:
@NickCPDX , have a look at https://github.com/TravelMapping/HighwayData/blob/master/updates.csv

It is sorted by region. Each region is sorted by date (newest first). Every new route, (relevant) realignment, extension or truncation needs an entry.

si404:
Something seems a bit janky with the I-580 file. Are all the points in twice? Certainly there's a stray half-line that will muck things up.

Otherwise, they seem fine (though I think US50Bus ought to have a US50 label, rather than a US395 one)

si404:
I've fixed I-580, which did repeat (copy-paste error).

PR #1585, opened 2017-09-05
Update updates.csv

NickCPDX:
Nevada updates


si404:
I think a bit more detail would help - point labels at either end of where extensions and reroutings occur would be good as a minimum. "rerouted to I-580" ought to be "rerouted _onto_ I-580 between x and y"and "extended to I-5**4**80" ought to be "extended from x to I-580".

ideally with extensions you would say which end (north, south, east, west) it is extending, and with reroutings describe the route it was on before, but those things aren't that useful. Also these updates aren't at the top of Nevada's listings, which doesn't matter for anything but anal retentiveness, but is an easy fix.

jteresco:
I'll hold off on this one and #1583 for tonight so we can get the updates in good shape.  I was never very good at making good updates entries, myself...

michihdeu:
onto to?

jteresco:
Are we good with the updates entries now?

PR #1588, opened 2017-09-06
datacheck and activations

si404:

PR #1592, opened 2017-09-07
fixes for #1588

si404:


si404:
I forgot to change folders of the asiah files (and while I checked the .csvs, it seems I didn't do that fully either!). Took me about 2 minutes to fix, and here it is.

PR #1596, opened 2017-09-09
Update nv.i515.wpt

NickCPDX:
I-515 decommissioned from 546/215 south; renumbered I-11.

### DIFF
--- a/hwy_data/NV/usai/nv.i515.wpt
+++ b/hwy_data/NV/usai/nv.i515.wpt
@@ -1,7 +1,3 @@
-56 http://www.openstreetmap.org/?lat=35.995369&lon=-114.934070
-57 http://www.openstreetmap.org/?lat=35.999952&lon=-114.963212
-+X01 http://www.openstreetmap.org/?lat=36.002015&lon=-114.979925
-59 http://www.openstreetmap.org/?lat=36.011825&lon=-114.989991
 61 http://www.openstreetmap.org/?lat=36.033518&lon=-115.014238
 62 http://www.openstreetmap.org/?lat=36.046011&lon=-115.022650
 64A +64 http://www.openstreetmap.org/?lat=36.063740&lon=-115.031791


rickmastfan67:
Is this being based on OSM?

If so, don't trust that right now, as the same user keeps adding it has been blocked a few times there for adding fictional data.

NickCPDX:
No. I have been correcting that user a lot!

The article below indicates that signs are in the process of being installed, and that 515 has been decommissioned south of 215. I wasn't going to believe it until I saw boots-on-the-ground evidence, but the quotes from NDOT look good enough. I also found a NDOT signing plan online during an image search today. 

http://m.lasvegassun.com/news/2017/sep/09/more-interstate-11-signs-on-the-way/


Nick Christensen
via iPhone (and I blame Apple's programmers for any typos)

jteresco:
Holding off on these for tonight, since I think the wpts are updated but not the csvs.  This will also give us a chance to make sure there's general agreement that it's OK to make this I-11.  I expect it's good to add, but others are better keepers of our rules for what new stuff gets included and when than I am.

jteresco:
I'd like someone better at making sure AltNames are inserted to minimize or ideally eliminate list breakage for this to take a crack at the usai and usaif csv entries for this.

NickCPDX:
Sorry, just to clarify, what needs to be done now?

michihdeu:
I don't think that AltNames are missing in the wpt files. Not sure if AltNames in csv files would work though...

I would use "New" instead of "Created" in updates.csv (which has conflicts now):
> 2017-09-09;(USA) Nevada;I-11;nv.i011.wpt;created route
->
> 2017-09-09;(USA) Nevada;I-11;nv.i011.wpt;New route

Is the eastern segment of I-11 in service? It's not yet on OSM nor GM.

jteresco:
OK, I have a few extra minutes tonight, so I'm going to pull these in and make the csv edits.

yakra:
> I'd like someone better than me at making sure AltNames are inserted to minimize or ideally eliminate list breakage for this to take a crack at the usai and usaif csv entries for this.

Just saw this now.
Taking a look...

yakra:
`NV I-11FutHen` is only in use in ssgtcrusty.list, whopperman.list, and sbeaver44.list.
All of them have current I-11 showing up in their mapviews, probably due to concurrency detection.
No work needs to be done here.

PR #1602, opened 2017-09-12
Fixing US 395

Bickendan:
Eliminates the doubleback and retrace of the southern half of the route.


jteresco:
OK, good.  Does this restore points in use as alt labels as well?

Bickendan:
US 395 had no affected points needing alt labels. 
Looking over OR 99 and 99W I'm not seeing any points that were badly affected by point name swaps in the pointsinuse log outside of Mapcat's (and they look to have been swapped on his end).

jteresco:
OK, great. I'll get this in tonight's update.

PR #1608, opened 2017-09-14
US 101, OR 58 cleanup, OR 39 Bus -> 39 Kla, US 97 Bus Kla, uploading first set of ORH

Bickendan:
'ORH' directory should probably be 'usaorh'
Invalid in-use point FaiMouRd on US101 deleted! 
Bus39Kla -> OR39Kla with reroute
Devel System ORH uploaded with first set of highways


jteresco:
Looking at the conflicts from this pull request, I'm worried that it was made without pulling down the latest from the origin on GitHub first.

Also, it won't work if the directory name with the files for the new devel system doesn't match its entries in the csv files.  Let's get that fixed up before I pull in.

Bickendan:
Since there was a lot going on with this pull request, I probably should have split it into two -- one for the updates, and one adding ORH. 
To resolve the conflicts, I should pull down updated copies of the .csvs from Master into my local repository, I'm guessing? 
And how do I fix the directory name ORH to usaorh?

ovoss:
In case this isn't being fixed already, along with other fixes: I noticed one revised route file for OR 58, as well as some new ORH route files, including waypoints referring to ORH routes. Should there be any references to unsigned routes in active/preview route files, or even in devel ORH files? I recall this issue coming up when I re-did the Alaska routes, and Tim told me to use waypoint labels for intersections with Alaska's unsigned Interstates (including where those Interstates intersect each other) that used the signed Alaska state route numbers, rather than the unsigned Interstate numbers.

If the ORH routes are signed at those locations, why not move those ORH routes into the main Oregon state route set?

jteresco:
I think it's best to abandon this pull request, merge in the latest from GitHub into your copy, and then do some smaller requests.

As far as the renaming, if it's not yet in git you can just rename.  Otherwise a "git mv" at the command line should do it.

jteresco:
It looks to me like the consensus is that the ORH routes should not be included at this time.  I'm going to close this one.  @Bickendan, could you create a new pull request with the other items that we do want to get in?

PR #1609, opened 2017-09-14
BEL/SK

michihdeu:


si404:
you need to rename the cansph .csvs to cansk else it won't work.

michihdeu:
Thanks!

PR #1611, opened 2017-09-16
Englan near miss points

si404:


michihdeu:
@si404 updates.csv issues:

- Date format wrong ("/" instead of "-")
- A361 (Frome) description is incomplete

PR #1615, opened 2017-09-20
Merge fixes

si404:


michihdeu:
fra.n007.wpt has incorrect OSM link (w/o "/")
I'm not sure that updates.csv has vaild format. It seems that you've removed the wpt file names including the whole cell (if it's really intended to be w/o file name, it should be ";;")

PR #1631, opened 2017-09-26
Update updates.csv

NickCPDX:
Updated NV 439 and US 50


yakra:
WRT the extension of NV 439, best practice would be to include a mention of both the old and new endpoints.

PR #1636, opened 2017-09-29
Update co.co009.wpt

compdude:
Move shaping point X03 to reflect a realignment of part of CO9 according to OSM.


ovoss:
Can two new labeled points be added in the file (instead of the relocated shaping point), one for where the new alignment peels away from the old one, and the other where the new and old alignments rejoin? These points should also be mentioned in an Updates item, for people who drove the old but not new alignments who want to remove the realigned segment from their list files.

compdude:
@ovoss: Done!

PR #1641, opened 2017-10-06
Spain regionalisation

si404:


michihdeu:
Please merge it so that I can add my stuff (corrections). Thanks!

michihdeu:
Jim has removed gbna from systems.csv. I that correct? https://github.com/TravelMapping/HighwayData/commit/e3eccf0466fa9726a924010631aed71f700fc5a3

michihdeu:
Sorry for bothering again but systemupdates.csv "Re-entered" is odd. It was never removed it's only reorganized...

si404:
> Jim has removed gbnb from systems.csv. I that correct?

Now you've re-read it it is. I took the Scottish routes and put them into an sctb holding system.

>Sorry for bothering again but systemupdates.csv "Re-entered" is odd. It was never removed it's only reorganized...

They are re-entered. Effectively the systems as-was were removed and replaced by reorganised systems. Given .list file entries need to be re-entered, the systems are. I'm easy with a change to reorganised, but that sounds like sorting routes differently, rather than breaking them up with none of the old ones working.

michihdeu:
systemupdates.csv syntax is wrong though (2x ";"):
2017-10-06;Germany;deua;Re-entered
-->
2017-10-06;Germany;deua**;;**Re-entered

PR #1653, opened 2017-10-10
AR Cleanup Part 9

yakra:
http://tm.teresco.org/forum/index.php?topic=8.msg8008#msg8008


yakra:
need to fix updates.csv per @michihdeu's comment on 8dd95d1

yakra:
Fixed.

PR #1661, opened 2017-10-14
fixes

si404:


jteresco:
espcm.n320 causes the site update to fail because of a duplicated line.  There seems also to be a blank line and a point that seems unlikely to be intentionally included (CM/MD).  Since I don't know the intent here, I'm going to comment this out of the csvs pending a fix.

PR #1665, opened 2017-10-18
PA372: recenter HolRd_E; add FriDr

yakra:
http://tm.teresco.org/forum/index.php?topic=1935


yakra:
will require a datacheck fix, probably a *nearmatchfp*.

PR #1674, opened 2017-10-26
Delete sct..wpt

yakra:
ping @si404 I can't imagine there'd be any objections here, right?


si404:
Any reason why you want to delete it? No objections from me other than I'm not sure what your objection is to it existing!

jteresco:
@si404 I think I'd prefer if all such files could be kept out of GitHub altogether.  I know they can be deleted easily enough later, but they can also be kept in your clone/fork as untracked files.  That said, I never had a strong enough objection to them being here until I saw this pull request, so if they help make your workflow more efficient, forget I said anything.

yakra:
This was a 0 byte file, right? I assumed it was erroneous.
Interesting that the [files tab](https://github.com/TravelMapping/HighwayData/pull/1674/files) for this PR says "No changes." (A GitHub bug for 0 byte files?)
It was shown on http://tmtest.teresco.org/logs/unprocessedwpts.log but now it's not.

PR #1746, opened 2017-12-03
islth bugfix

Duke87ofST:
Fixed error where TH255 was not in csv files


yakra:
ping @seaweedfactory -- Iceland Þjóðvegur = an active highway system

PR #1815, opened 2018-01-08
Minor CA updates

ovoss:
Minor tweaks to active US 101 route file, per forum discussion, and preview CA 19 route file.


jteresco:
Network trouble has my whole domain offline.  I'll get this in as soon as we can get the server back on the network.

PR #1816, opened 2018-01-10
NW

michihdeu:


jteresco:
After a quick investigation of these errors, I realized I'm not sure what to fix, so I had to comment out deub to get tonight's update to go.

`ABORTING due to 3 errors:
1: Duplicate root in con_route lists: deurp.b054
2: Check FAILED: 39072 routes != 39071 connected route roots.
3: route deunw.b054 not matched by any connected route root.
`

PR #1823, opened 2018-01-14
canab fixes for AB8xx routes

yakra:


yakra:
I may add some more commits to this.
No problem if it gets merged in before I do so, though.

jteresco:
I'll probably run the site update tonight around 10 PM EST.

yakra:
> No problem if it gets merged in before I do so, though.
Actually it'd be a problem if it's merged after I commit either AB16 or TCHYel, but before I commit the other. This should be ready in a couple minutes.

yakra:
Ready to rock.

PR #1847, opened 2018-01-26
Create NV 529.

NickCPDX:
SR529 - Carson St. - RMID #114 - S-N
Jct of US50/US395/SR529 to the south R/W line of Fairview Dr.


jteresco:
I believe this will need an updates.csv entry as well.

michihdeu:
Plus usanv.csv and usanv_con.csv

PR #1860, opened 2018-02-01
Activate usaak (part 2)

ovoss:
Jim, does this take care of all I need to do to move usaak to active status?


ovoss:
yakra just asked on the forum that activation wait a few days, to give him a last shot at comments,

PR #1862, opened 2018-02-03
OH minor changes; AR1-AR14

mapcat:


jteresco:
File names did not match new names in usaoh.csv so I changed them.  Wasn't 100% sure that I shouldn't be changing the csv entries instead.

mapcat:
Yes the .csvs had the correct names. Sorry I was distracted when I
submitted those. Thanks for making the changes.

PR #1880, opened 2018-02-11
fix minor issues with engb

si404:


jteresco:
This undid the fixes I made with the extraneous entries in engb_con.csv.  I'm not sure if I did something different or wrong or if you did, but I'll make the fix again and hope for the best.

si404:
I merged from tm/master after having added some missing lines and hadn't realised there were extraneous lines removed. if you take them out again, it should work fine.

jteresco:
Looks like it's good again.  All the lines in the file had a region field, which isn't included in the _con.csv files (as I know you know but assume you just didn't remember to remove when creating that file from the .csv).

PR #1922, opened 2018-03-23
AK7Pet tweaks

ovoss:
Minor changes to route file, mainly to relabel side road to now-defunct Rainforest Islands ferry. Relabeled points are not in use.

With these changes, and completion of yakra's review, usaak is ready to activate (per earlier pull requests, except new system updates table entry needs new date).


jteresco:
OK, great!  I'll make the edits to the systems and systemupdates files for this before tonight's update.

PR #1953, opened 2018-04-11
Update ms.i055.wpt

ajfroggie:
Correction for a missing waypoint along the I-20/55 concurrency in Jackson.

### DIFF
--- a/hwy_data/MS/usai/ms.i055.wpt
+++ b/hwy_data/MS/usai/ms.i055.wpt
@@ -28,9 +28,10 @@ LA/MS +0 http://www.openstreetmap.org/?lat=30.999792&lon=-90.505669
 90A http://www.openstreetmap.org/?lat=32.242062&lon=-90.214834
 90B http://www.openstreetmap.org/?lat=32.252733&lon=-90.211358
 92A http://www.openstreetmap.org/?lat=32.268193&lon=-90.209942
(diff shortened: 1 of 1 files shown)


yakra:
880d15f745d2fc880652dc99e9994d160fb79e47: why the label change? I-55 is an exit-numbered route, so 44(20) etc. are the correct syntax.

ajfroggie:
http://tm.teresco.org/forum/index.php?topic=1793.0

yakra:
That doesn't explain the label syntax change though.

ajfroggie:
This is how the MS lists have always been, and before I realized Tim had implemented a change to CHM labeling.  At this point, given their use, I'm just going to leave it as-is.

yakra:
> This is how the MS lists have always been

[It's not though.](https://github.com/TravelMapping/HighwayData/blob/master/chm_final/usai/ms.i055.wpt#L31-L33) If it had been, there'd be nothing to change!
Tim didn't change anything; Interstates have always used ["Interchanges on exit-numbered highways"](http://travelmapping.net/devel/manual/wayptlabels.php) style labeling.
Canonical exmple: [IA I-35](https://github.com/TravelMapping/HighwayData/blob/b1e1296c93fd22dbad7107ae3e3f96cadb45da6d/hwy_data/IA/usai/ia.i035.wpt#L20-L28)

PR #1966, opened 2018-04-21
OH fix + more usaar

mapcat:


jteresco:
I made a bunch of fixes to the csv files that I will commit and push.  Mostly, there were capital letters in filename specifiers, which works when I run an update on my Mac with its annoying case preserving but not case sensitive filesystem, but fails when I run an update on FreeBSD (where I intend to move all updates within a few weeks).  Also, there were some names with spaces "el " where you want "eld" which are not going to work on the Mac or FreeBSD.  Please pull and merge in my fixes before making any changes to your local files.

jteresco:
Also I am removing these lines, as there are no matching wpt files:

```
usaar;AR;AR4S;;;;ar.ar004s;

usaar;AR;AR7S;;mar;Marble Falls;ar.ar007smar;

```

mapcat:
Sorry, I used an automated process in Excel to create the csvs and failed to notice that it caused those errors. I'll fix my formula. Thanks for letting me know!

PR #1997, opened 2018-05-25
Iraq/Jordan boundary

yakra:


si404:
If you are going through the borders might it make sense to rename them so it is obvious which file is which? Also I think I lost my do Ireland properly that took ages to make as I thought it wouldn't be needed 😥

jteresco:
As I understand it, @yakra is updating boundaries for a side project.  At this time, they are not used in any way by TM.  If we were ever to generate our own maps like CHM did, the boundaries would be used.

michihdeu:
It is not used by TM but shown on TM's datacheck.php... (even in red)

yakra:
> shown on TM's datacheck.php... (even in red)

I don't see boundaries there at all. Shouldn't be, if boundaries aren't even included in systems.csv. 

> As I understand it, @yakra is updating boundaries for a side project. At this time, they are not used in any way by TM. If we were ever to generate our own maps like CHM did, the boundaries would be used.

Right. Not used in any way by TM, but I've been working on [CHM-style Javascript maps](https://github.com/TravelMapping/Web/issues/29) as a side project. Cleanup of the boundaries is involved, but is a very low priority. This commit was just a way to kill a little time before work while the YouTube video I was listening to in another browser window didn't require my full attention.

michihdeu:
Maybe I'm wrong but I've never seen this before:

http://tm.teresco.org/devel/datacheck.php?sys=null&rg=MT

`mt.sr320 | MainSt | OUT_OF_BOUNDS | (46.500704,255.20695) | mt.sr320;MainSt;;;OUT_OF_BOUNDS;(46.500704,255.20695)`

michihdeu:
Si - I know..................................... I only pointed out that anything with bounds / boundaries IS shown on TM data check. Period.

si404:
I think you are having a non-native speaker of English problem here.

Nothing to do with the boundaries are on datacheck. The idea that 3 'out of bounds' errors that have nothing to do with the boundaries now are listed doesn't mean that 'anything with bounds/boundaries' is on datacheck (which implies any error we look for in boundary files) - it does mean something to do with bounds/boundaries are on datacheck, of course. But that something has nothing to do with the boundaries we're talking about with this pull request.

michihdeu:
Maybe, sorry.

PR #2008, opened 2018-06-01
KY, OH, & MI fixes

mapcat:
Jim: I did not have time to set up an account to test this, so it is untested (I need to submit today since it will be another 2 weeks before I can work on this again). However, I have not made any additions to usaar.csv or usaar_con.csv, which are the only files that have caused trouble in the past.


yakra:
I downloaded mapcat:master and test on my machine.
> Data check successful!

mapcat:
@yakra: thanks!

PR #2022, opened 2018-06-15
usaar -> preview

mapcat:


mapcat:
Jim, the hyphenated routes (e.g. AR949-4) aren't working. What do I need to do to those?

jteresco:
I guess we've never had hyphens in list names before, so something doesn't like that.  I think they're successfully in the DB, but the web front end isn't happy in some cases.  I'll have to check it out.

mapcat:
Yes, the one I clinched shows up as 100% and looks fine on the map, so the DB doesn't seem to mind them.

jteresco:
Well, here it is:

https://github.com/TravelMapping/Web/blob/fa645deed1ec07164c4a4b8515a3f6deb4151c07/lib/tmphpfuncs.php#L399

Hyphens are specifically disallowed in filenames.  Any harm from your view in renaming the files not to have the hyphens?  They could remain in the .list names, so the change would not be visible to users.

mapcat:
Sure, but will they sort correctly if I remove them? I wouldn't want 9021A after 980, for instance.

PR #2023, opened 2018-06-16
Remove hyphens from usaar roots

mapcat:


jteresco:
Running an extra site update this morning with these.

mapcat:
Thanks--looks like everything is working, and the order in the table was not affected.

PR #2042, opened 2018-06-26
Downtown Buffalo cleanup

yakra:
buffalo256
This reminds me of all the fun I had in Watertown with #7 a couple years back...

## Changes:

**NY5:**
* Complicated route downtown... Couplet:
* Eastbound, from Buffalo Skyway: Church, S Division, Ellicott, Goodell, to Main.
* Westbound, from Main: Pearl, Chippewa, Elmwood, to Buffalo Skyway.
* Changed to just cut the diagonal across the couplet, with a route intersection point at each end.
* [NY16/384](http://www.openstreetmap.org/?lat=42.884015&lon=-78.879690) -> [NY384](http://www.openstreetmap.org/?lat=42.883722&lon=-78.878971):
  - 16 doesn't actually intersect here (see below).
  - Coords tweaked to have a graph connection with NY384.
* `NY266/354` eliminated, folded into `NY384` (was in use by *hoopitypoop* & *keithcavey*). 
* `NY33` left in place; see below.

**NY16:**
* Inaccurate west end; there's no multiplex along NY5/Ellicott to end near NY384.
* The actual terminus is at [Ellicott & S Division](http://www.openstreetmap.org/?lat=42.882328&lon=-78.873119), where eastbound NY5 makes a 90° turn.
* Northbound end: via Swan St & Elicott St, to S Division.
* Southbound beginning: via S Division, Oak, and Seneca.

**NY33:**
* New endpoint: The western terminus is at Ellicott St, eastbound NY5.
* Eastbound beginning via Tupper St.
* Westbound end via Goodell St.
* Normally, the sensible thing to do would be to put a point midway between the two, but what NY5 is doing here gums things up a bit.
* Hence I've fudged the endpoint a bit to meet up with existing `NY NY5 NY33`, and have a graph connection.
* `HamSt` added, for connection to I-190(5).

**NY266:**
* Truncated to `I-190(8)`, which is a small tweak to the existing `VirSt` point.

**NY354:**
* [NY5/266](http://www.openstreetmap.org/?lat=42.886121&lon=-78.875613) moved to [NY5](http://www.openstreetmap.org/?lat=42.885881&lon=-78.871957), at Ellicott St, eastbound NY5.
* `WilSt_E` and `CliSt_W` added to clarify routing.
* `I-190` *(new graph connection!)* and `OgdSt` (to I-90 @ 1) added

**NY384:**
* `NY5/16` +-> `NY5`
* `NiaSq` added, for connections to the beginnings of NY266 and NY354. Think of it as taking the place, graph-wise, of the old `NY5/NY266/NY354` point, only with none of those routes actually connecting here.
* `ChiSt` added, for westbound NY5.
* `EdwSt` added, for connections to/from I-190(8)/NY266 in the west, and NY33 in the east.

## Graph connections:

* **NY16** now has a hanging N end @ EB NY5, while NY5 itself cuts diagonally across the couplet.
* **NY33** now has a graph connection to NY5; no more hanging end.
* **NY266** now has a hanging S end at the end of the I-190(8) interchange. Unavoidable here, as this is in fact the route's end.
* **NY354** now has a hanging W end @ EB NY5, while NY5 itself cuts diagonally across the couplet.

@jteresco, if you'd like to keep the graph connections with NY16 & NY354, I could be convinced to pull a [MA US202](http://forum.travelmapping.net/index.php?topic=2200) and route NY5 along one side (EB) of the couplet. What do you think?


jteresco:
Let's see how the graph looks.  I'd like to keep things connected if they're connected, but don't want to do so at the expense of making the TM side less "good" in some sense.

Your post brings me back to the trouble I had in that area when drafting usany way back.

yakra:
### Now:
buffalo-diag256
---
### "MA US202" style:
buffalo-east256

yakra:
@jteresco, any thoughts as to how you'd like to see this handled? I'm willing to change NY5 to the example shown in the last graph.

jteresco:
It would make sense to me that they should connect.  I just looked in the Traffic Data Report and NY 16 and NY 354 are shown as meeting NY 5.  The "Potential Future" looks good to me.

PR #2081, opened 2018-07-30
Europe changes

si404:


jteresco:
Since I'm traveling and can't easily make fixes if any are needed, I'll wait until the first update after I'm home to pull this one in, unless you've had the chance to run it through the datacheck.sh program.

michihdeu:
I think it's not good that update entries dated 2018-07-20 appear on 2018-08-01. 2 or 3 days are fine but....

S8zam entry is incomplete: "Route extended at west end from"

PR #2082, opened 2018-07-30
LA point requests & shaping points; TN exit & NMP

mapcat:


mapcat:
Data check successful

mapcat:
Thanks...I'll probably wait until close to activation to do a final check with HDX.

PR #2103, opened 2018-08-10
usafl: FL-9B new extention to CR2209

rickmastfan67:
https://www.news4jax.com/traffic/northbound-lanes-of-state-road-9b-open-in-st-johns-county?__vfz=medium%3Dsharebar

Not sure if the I-95 will now gain an exit number, but I'll cross that bridge once somebody checks out the new alignment.

### DIFF
--- a/hwy_data/FL/usafl/fl.fl009bjac.wpt
+++ b/hwy_data/FL/usafl/fl.fl009bjac.wpt
@@ -2,6 +2,6 @@ I-295 http://www.openstreetmap.org/?lat=30.180469&lon=-81.513609
 +1 http://www.openstreetmap.org/?lat=30.165675&lon=-81.511163
 2 http://www.openstreetmap.org/?lat=30.134805&lon=-81.509318
 3 http://www.openstreetmap.org/?lat=30.130848&lon=-81.510240
-I-95 http://www.openstreetmap.org/?lat=30.114617&lon=-81.514156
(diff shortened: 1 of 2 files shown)


rickmastfan67:
Also, not sure if this part will become I-795 in the future, so, not extending that file.  But I have a feeling that I-795 will just be the I-95 -> I-295 segment.

rickmastfan67:
NM, just found a video that somebody took back on 05/21/17 that clearly shows an exit number for I-95 painted on the ground (no gore signage yet).  It's Exit 4.  https://youtu.be/qVplSYgmnRs?t=2m13s

I'm sure that FDOT updated the signage since then and posted it for all to see, so I'll edit both FL-9B's & I-795 (Future)'s files to have I-95 listed as '4'.

jteresco:
Should 9B now be just "FL9B" instead of "FL9BJac"?

jteresco:
It doesn't look like we're consistent elsewhere.  The first two I looked at were NY and MA, both of which I initially developed but relinquished a few years back.  usany doesn't include a city or suffix for any except NY120A (which has 4 segments) while usama includes them for every route with a letter suffix.

rickmastfan67:
Well, as I said above, I can make FL consistent (I know I did a few without the name without realizing it) in a separate changeset later since usafl is still in 'preview' and would not have to push an 'update' log entry for the changelog file.

PR #2110, opened 2018-08-14
Joisey

yakra:
nj.i080;I-80_U;;;LABEL_SELFREF;


yakra:
I'll bundle this with some *canabs* FPs in another branch instead.

yakra:
Derp. No I won't. Those are all VISIBLE_DISTANCE; no need to add to FP list.
This pull request is back on/

PR #2116, opened 2018-08-14
New England point requests from forum

yakra:
http://forum.travelmapping.net/index.php?topic=2590


yakra:
http://forum.travelmapping.net/index.php?topic=2591

PR #2130, opened 2018-08-19
Utah updates/datacheck fixes 2018-08-19

Duke87ofST:


jteresco:
@Duke87ofST Let's get you set up to be able to run the datacheck before pull requests to avoid failed updates.

PR #2139, opened 2018-08-22
AB58 relocation @ AB35, High Level

yakra:
http://forum.travelmapping.net/index.php?topic=2603


yakra:
Nota bene: these commits are also included in #2140

PR #2141, opened 2018-08-22
PA point fixes from forum

yakra:
http://forum.travelmapping.net/index.php?topic=2608


yakra:
http://forum.travelmapping.net/index.php?topic=2611

PR #2145, opened 2018-08-24
datacheckfps: WV FPS

rickmastfan67:
Clearing two FPs for WV on that list.  Not adding WV-97's one till I get to it on the usawv overhaul, as I think the angle % might change, or labels involved, once I look into that route some more.


rickmastfan67:
Also just quickly threw in an error I noticed on the 'updates.csv' file where an entry for UT-190 was lacking the 'ut' part for the file route, leading to a broken link to that route.

PR #2146, opened 2018-08-24
BY/TH

michihdeu:


michihdeu:
@jteresco I didn't find the csv issue by hwy data check but in my user log file only because I've traveled B87 and got "wps not found". I've renamed b87apo to b87 and b87 to b87kra but forgot "abbrev" change only.

jteresco:
That sounds like something that might be useful to check for.  Could you put together an issue with the example and I'll take a look when I have a chance?

PR #2147, opened 2018-08-24
update updates.csv

ovoss:
Redo of update items re: Boulder City NV


jteresco:
Looks like this one left the encodings alone.  Thanks for the do over.

rickmastfan67:
@ovoss, I wouldn't be adding '(NONE)' in the route file name section, as that adds a link there.  Just leaving it blank in that part between the ';' automaticly adds the '(NONE)' text. 

See the "I-15BL (Brigham City)" (Utah) entry on 2018-08-11 as an example.

So, I'd recommend going back and removing the '(NONE)' text in the updates file.

PR #2168, opened 2018-08-29
Adjustment of south endpoint of ND 67

the-spui-ninja:
I noticed that ND 67 was a bit off (it goes over the railroad tracks on a bridge at the south end), so the adjustment was made.

If this causes any problems let me know (I figured a small thing would be good for my first HighwayData pull request).

Source: http://mapserver.co.bowman.nd.us/


jteresco:
I don't see any problems with the pull request and it will be live on site in 30 minutes or so.  Thanks!

PR #2176, opened 2018-09-01
Restore my fixes to recent Nevada entries

ovoss:
I removed (NONE) from a field in three recent Nevada updates items. My guess is the removals were undone by someone editing an old version of updates.csv.


jteresco:
Shouldn't be a problem unless you and @michihdeu both made edits to the same part of the same file.  The beauty of source code control.

ovoss:
Jim:


We both made exactly the same change to the same three lines.


It seems that our changes made it into the master without trouble.


Oscar

PR #2194, opened 2018-09-04
AZ NMPs @ state line

yakra:
Ping @si404 again with a usaush edit.
No action required from @ovoss; his coords at state lines were all more accurate, and replaced those in the AZ files.

This should leave `AZ I-15 AZ/UT` as the only remaining AZ NMP. I recommend changing this in both files: http://www.openstreetmap.org/?lat=37.000080&lon=-113.623360
Ping @Duke87ofST.


Duke87ofST:
I-15 NMP is fixed in https://github.com/TravelMapping/HighwayData/pull/2196

PR #2225, opened 2018-09-13
American Samoa nmpfps

yakra:
Every [NMP in AS](http://travelmapping.net/logs/nmpbyregion/as.nmp) is a FP, so I added them to the list.
AS is one of @si404's regions, and I did this without asking permission, so let's get his OK before merging this in.

@si404, while you're here: Everything in [HUN](http://travelmapping.net/logs/nmpbyregion/hun.nmp) also looks like a FP, including one "LI" pair that looks like it was done to avoid an M85/F85 FP concurrency to the west. Everything in HUN can also be marked FP, right?


michihdeu:
@yakra just a quick note: I'll check my regions anytime this fall. But I wanna do it because it's always a good "general check" for issue. Thanks.

yakra:
I don't plan on doing any more NMPs in anyone else's regions. I was having some fun with filtering the false positives out of the nmpbyregion files & seeing what the smallest size files left over were. These were a couple simple cases I happened to come across.

PR #2226, opened 2018-09-14
GA: US 1 bypass of Oak Park

the-spui-ninja:
See http://forum.travelmapping.net/index.php?topic=23.msg68#msg68

I think I have it right as per http://www.dot.ga.gov/DriveSmart/MapsData/Documents/CountyMaps/Emanuel.pdf and http://www.dot.ga.gov/DriveSmart/MapsData/Documents/CityMaps/OakPark_EmanuelCounty.pdf, but by all means double-check what I did.


the-spui-ninja:
Should this go in the updates file?

jteresco:
Yes, an updates entry for the reroute of US 1 should be included, describing the change pretty specifically.

jteresco:
Also, the GA routes are in preview, so no updates entries are needed for changes to those.

PR #2275, opened 2018-10-09
AL/TN border point reset

ajfroggie:
Reviewing the AL/TN border points and adjusting as needed based on signage and pavement changes.  Not all border points needed adjustment.  Also revised the border points in Ardmore as all three routes there intersect AL 53 along the state line...no labels were changed but I did center the points in the relevant intersections.


rickmastfan67:
@mapcat

PR #2280, opened 2018-10-10
FL/AL border point recentering

ajfroggie:
Recentered points along the FL/AL border based on signage/pavement locations and aerial imagery.


rickmastfan67:
Literately just found this pull after noticing some errors in a pull request I was putting together for usafl today. @ajfroggie, please next time tag me if you're going to edit a 'preview' system that I'm working on, as I manually keep my files updated while working on a system like this.  Thanks. :)

PR #2320, opened 2018-10-28
Utah NMP fixes and UT290 deletion

Duke87ofST:
UT 222 is also technically truncated in this pull. This is not mentioned in updates.csv since truncation does not remove any waypoints and no action will be required on the part of any users - waypoint for northern terminus was relocated from the arbitrary point along the road where it was to the correct terminus of route about one mile south per HRO.


jteresco:
Please try to run the datacheck.sh before submitting pull requests to catch errors.

PR #2337, opened 2018-11-03
deterministic HIDDEN_JUNCTION & VISIBLE_HIDDEN_COLOC

yakra:
I want to be sure @michihdeu & @si404 are OK with these changes.


jteresco:
So I'll hold off merging for now.

yakra:
> Why are these wps not visible on HDX? Maybe it's intended but I think it should be shown by default and not being dealt as "hidden markers".

Limitation of the HDX + the TMG format.
The TMG format has no real way of storing whether a vertex *(that is, the points representing each **end** of an edge, rather than the always-hidden shaping points along a collapsed edge itself)* is hidden or not.
It looks like (I've never gotten under the hood of the HDX) the HDX determines whether a vertex is hidden by the presence of a `+` in its label (Either a `+` after an `@`, or a `+` anywhere, I don't know).
Thus  it knows which points to hide, particularly when displaying simple graphs.

Depending on how forming a point's `unique_name` works out, a VISIBLE_HIDDEN_COLOC can sometime end up with a `+` in the `unique_name`, and sometimes not:
• aut.a007@S10 + aut.e55@+A7/S10 + aut.s010@A7 = A7@S10&S10@A7&E55@+A7/S10
• chngx.g65@G72_E + chngx.g72@+G65/G76 = G65/G72@G72_E

Seems to me the thing to do, rather than rework the TMG format & the HDX, would be to, during graph generation, when a VISIBLE_HIDDEN_COLOC or HIDDEN_JUNCTION is detected, make sure that its `unique_name` doesn't contain a `+`.

`unique_name` already exists when these datachecks happen. Do we modify an existing `unique_name`, or make this happen when `unique_name` is initially created?

I haven't dug into how `unique_name`s are constructed, simplified, and passed around much yet.

yakra:
https://github.com/TravelMapping/EduTools/issues/223

michihdeu:
The NOR entry is also fine. Can you please merge it.... _I would not expect further objections...._

yakra:
ping @si404 again

PR #2353, opened 2018-11-10
poldw8

si404:


michihdeu:
fra.nvs exceeds limits b/n D117 and D902_E
systemupdates.csv entry should be 800-899 instead of 600-699

si404:
Yes, that was the order that, because it was upheld at great human cost, is why it is the 'Sacred Road' for the French.

PR #2359, opened 2018-11-12
Remainder of the AL state routes

ajfroggie:
Adding the remainder of the AL state routes and updating the systems .csv's for usaal.  Not sure what the 'Merge pull request #1" from November 1 is doing in here...that was my attempt to update my local fork.

As for the AL routes...unless I'm missing something, this should be enough to move the usaal system to preview.


jteresco:
Would you like me to move usaal to preview tonight or would you prefer to see the full set in devel first?

jteresco:
We now have duplicate routes between usansf and usaal:

<pre>
ERROR: Duplicate list name: AL AL152
ERROR: Duplicate list name: AL AL158
ERROR: Duplicate list name: AL AL255
ERROR: Duplicate list name: AL AL759
</pre>

It looks like our usual procedure has been to leave these in usansf until the new system is activated to avoid demoting an active route to preview.  That's what I'll do for now.  I'll also use the new usaal route files, which are likely more up-to-date.  The exception is AL 255, which doesn't have a wpt file in usaal.

PR #2402, opened 2018-12-02
fix greece

si404:


si404:
This just fixes a copy-paste error, where I didn't make grceo a preview system, merely moved it to that part of the list. Merging straight away.

PR #2403, opened 2018-12-02
I-95: Removing Alternate Labels 46, 48, 51, and 999

Markkos1992:
See the following links:
US 219 thread: http://forum.travelmapping.net/index.php?topic=2758.msg12275#msg12275
I-95/I-295/I-276 thread:  http://forum.travelmapping.net/index.php?topic=2321.msg12291#msg12291

Users affected by removing of alternate labels:
46-sneezy, jpinyan, epzik8, moshitea, verumsolum, inagaddadavida, codyg1985, okroads, ntallyn
48-rlee
51-jpinyan, bhemphill
999-dnthrox, neilbert, aaroads

All of the users referenced above will have error in their respective logs for "PA I-95" that will need to be corrected.


rickmastfan67:
Looks good to me.

However @Markkos1992, I'd remove those spaces you introduced in the 'updates.csv' file between the file name and the description of the change (";pa.pa297; Route"), between the ";" & the word "Route".  That isn't needed.  While I don't think it will cause any issues for the update page, it also can't hurt to keep it status-quo like all of the other entries.

Markkos1992:
Done to be consistent.

PR #2428, opened 2018-12-19
Update README.md

rmihailucian:
Added the romanian geoportal as source. Seems to be pretty reliable.


jteresco:
If the new source is actually used, it should be included on the site on the credits page.

rmihailucian:
Until now, I haven't done any changes according to the new source, but we (me and michih) are planning a general overhaul of roudn, to correct some errors and make it prone for adding the 5-tier system (roudj) sometime in the near future.

PR #2432, opened 2018-12-20
Update roudn.csv

rmihailucian:
Corrected some incorrect city names.


jteresco:
@michihdeu do you agree on the changes here?

michihdeu:
http://cmap.m-plex.com/tools/manual_syshwylist.php

> The City should be spelled as the locals spell it and using international characters as needed. 

We had the English version of București and a wrong spelling of Brăila (lucian's hometown).
We support unicode and I don't think that "ă" and "ș" need to be replaced. We do also use special characters for other countries, e.g. nor.

@rmihailucian Thanks for your change but please use correct Romanian spelling. Just make another commit to the same file.

michihdeu:
@rmihailucian roudn**_con**.csv needs to be updated too.

michihdeu:
@rmihailucian _con.csv file?

rmihailucian:
Done also _con.csv file.

michihdeu:
thanks, saw it just some seconds later.

I've run a data check and both csv files are fine - do not crash siteupdate.py

PR #2460, opened 2018-12-30
ME ParkLpRd: remove Sieur de Monts Rd

yakra:
Since Maine is my state, I'm just going to go ahead and do this.
[Maine E911 Roads](https://geolibrary-maine.opendata.arcgis.com/datasets/maine-e911-roads) and its predecessor shapefile sets show Park Loop Rd as Park Loop Rd, and not "Ocean Drive" or "Jordan Pond Road" as shown in OSM. Sieur de Monts Rd is shown as such, pretty clearly a separate entity, so cutting it out.
AltLabels are used to avoid breaking .list files.
Ping @si404.
FWIW, there's still a graph connection to the outside world, via [Paradise Hill Rd](travelmapping.net/hb/?r=me.parhillrd).


si404:
Thanks for doing this. I'd personally much rather have region maintainers update and fix stuff themselves (and I won't do it for them when the system is active).

As usanp is in preview, updates aren't needed, but it's not like they aren't useful.

Merged.

PR #2473, opened 2019-01-01
TN, LA & AR minor changes

mapcat:
New truck route in TN
Delete LA3057, switch ends of LA 3100/LA3100Spr
Delete AR980 (Wynne)
Fix typos


mapcat:
IN: reroute US 50, eliminate IN 750
KY: short new KY route and extension

PR #2483, opened 2019-01-05
DNK

michihdeu:
I've finished drafting a single route and Si brought 8 (eight!) systems to preview at the same time....


si404:
You started that route last year, I drafted 9.5 systems in 3 days (and most of that was trying to research the SADC RTRN and failing to find a complete source) - slow coach! 😜 

Why do you think I didn't even make a start on the Margarite route - it's a nightmare! My systems were a doddle not many routes and fairly easy to draft. Especially those long routes through empty wilderness where you could do a 200km route in 5 minutes like that.

PR #2487, opened 2019-01-06
DNK

michihdeu:


michihdeu:
I found a new route (SR565) and need to run another data check. SR565 is confirmed by OSM and GSV from 2009 (west end) and August 2017 (east end) but missing on wikipedia.

PR #2492, opened 2019-01-08
UT point additions to Needles Scenic Drive

ovoss:
Per forum discussion.

@si404 @Duke87ofST  any comments, etc.?


Duke87ofST:
Looks good to me

PR #2495, opened 2019-01-11
Where the giraffes are, and the zebra

si404:


yakra:
What about lions?

michihdeu:
I guess Si did not see lions on GSV?

si404:
> What about lions?

[Got Lions and Tigers - only in Kenya,](https://www.youtube.com/watch?v=FbYtASAakA) ;P

> I guess Si did not see lions on GSV?

No GSV in Keyna, but you can see them in [this video](https://www.youtube.com/watch?v=qVjGrlGG1rs) (auf deutsche). ;P

PR #2502, opened 2019-01-13
Activate grceo

si404:
Activates grceo, previews 7 systems in East Africa, and usaas


si404:
Well the system isn't really Kenya, but the remnants of a colonial system that is mostly in Kenya, but covers Uganda (and used to cover Tanzania too).

Plus, there's some hints that maybe Uganda has got/is getting its own numbering system, so calling it kena allows me to simply drop the two Ugandan routes in kena (and the one in kenb) and have it as a Kenya-only system if and when that system becomes draftable.

PR #2503, opened 2019-01-13
DNK/MAR

michihdeu:


michihdeu:
@si404 The route does still have issues. Especially label names and position of hidden wps.

I also wrote to @rmihailucian that he needs to consider your mar.tah1.wpt file and that he should write you a PM

btw: He has already drafted ALL marn routes but they need to be revised with my notes. The next iteration will be commented on the forum

PR #2504, opened 2019-01-13
Mozambique

si404:


michihdeu:
@si404 Mauritius Motorways tier 3?

si404:
Yes, because while they are called 'Motorways' they are merely major highways.

michihdeu:
ok 😄

PR #2508, opened 2019-01-14
Update mar.a001.wpt

rmihailucian:
Repositioned N1_A to the roundabout and renamed R204 to N7A (Morocco did a renumbering of their highway system).


michihdeu:
Point in use, I'll fix it! @rmihailucian (see email) @si404

PR #2512, opened 2019-01-14
Create mar.n002.wpt

rmihailucian:


jteresco:
See if you can get these commits to go into a single pull request (in the future).  Other highway data contributors can hopefully share their workflows with you that will help.  The multiple pull requests will work, it just takes longer to merge them all in.  In any case, thanks for contributing!

rmihailucian:
Hi, I also wanted to ask how it is better, a pull request for each or a pull request after more commits. Is it also possible to directly upload these files? I tried, but received a message that I am not allowed to directly upload files.

PR #2545, opened 2019-01-20
MAR updates

rmihailucian:


michihdeu:
@rmihailucian alt labels are usually after the normal label

+Xxx Yyy
-->
Yyy +Xxx

PR #2553, opened 2019-01-23
marn devel -> preview

rmihailucian:


michihdeu:
@rmihailucian you made the same fault some of us often do 😄 
You've perfectly copied the line but forgot modifying it. devel --> preview in last column 🤣

rmihailucian:
And I was prepared to open a bottle of wine for my first preview system :) Corrected, now I hope everything is ok.

michihdeu:
Cheers! (I already had some wine today)

jteresco:
Good news!  Looks like things are in order so this will be on site tonight.

PR #2562, opened 2019-01-29
Updates for turo around Istanbul.

rmihailucian:
Overhaul of the motorway system around Istanbul (last updates here were done over 2 years ago). I updated the system with the new openings/renumberings.


michihdeu:
@rmihailucian are you sure that updates.csv #2563 is complete? 

You've renameed O3Cob to O7Cob but only mentioned that O7Cob was extended.
O7 and O7Aks are new but you only wrote 'Extended to current end point.'. It was not extended but renamed and extended and another new route.
You've also deleted O6Con (or merged into main route?)

You need to describe the changes complete and more detailed.
'to current end point' is also too less. Just check the csv file for examples or ask on the forum before changing.

E.g. 'extended from XXX to YYY' (wp names!)

rmihailucian:
@michihdeu I put the more important updates, there were quite some, and I didn't want to have the visible updates list full of the new changes. It was more or less an overhaul of the motorways around Istanbul, the last update in the area was more than 2 years ago.

michihdeu:
But it's an active system and you need to indicate all changes to active systems!

PR #2581, opened 2019-02-04
Move WA 99 into new tunnel

compdude:
The new SR 99 tunnel under downtown Seattle just opened today so I am updating its route to move it to the new tunnel. 

http://forum.travelmapping.net/index.php?topic=2869.msg12995#msg12995

Note: the PR changes aren't supposed to be in here; it's because I edited something for PR in my own branch that I didn't pull request a long time ago, and for some reason it still remained in my branch even though I thought I updated it to match what was on TM. The two updates for usapr.csv and usapr_con.csv cancel each other out as a result. 

Also, I'm not sure why it removed all those extra spaces in the updates.csv file. I used GitHub desktop for this update, and it seems like the Atom text editor did something weird with the updates.csv file.


jteresco:
I think the removal of the extraneous spaces is a positive thing, so no concerns there.  But what about the added KY 8 updates entry?  Where did that come from?

jteresco:
It looks like it's supposed to be there, so I'm merging in.

PR #2607, opened 2019-02-20
turd additions

rmihailucian:


si404:
When are these additions going to be added to .csvs?

rmihailucian:
I plan to add them after finishing drafting, while doing the final check. Should I add them before?

michihdeu:
It's up to you. Both is fine.

PR #2641, opened 2019-03-09
turd -> preview

rmihailucian:

PR #2643, opened 2019-03-10
Corrected errors from #2641

rmihailucian:
Corrected errors from #2641, now it should be ok. I would bring the system to preview in the next step, if everything goes well.


si404:
I'm going to check this - looks like the errors are fixed.

si404:
Yep, this works!

rmihailucian:
Perfect, then I will also bring it to preview.

si404:
That's already there with the previous pull - no need to do anything.

michihdeu:
No, it wasn't.

PR #2644, opened 2019-03-10
turd -> preview

rmihailucian:


michihdeu:
turd was commented that means it's NOT included in data process. You need to run data check AFTER you've included turd to systems.csv.

si404:
I ran datacheck having unchecked it locally.

I'm going to close this as it's just going to be easier to do that to avoid conflicts with my pull request I'm working on.

PR #2671, opened 2019-03-16
The country that lost a war to birds

si404:


yakra:
"What... do you call this place?"
"It's a barbecue area!"

PR #2722, opened 2019-04-09
Delete ak.ak010chi, cansk edits

ovoss:
Both deletion of AK 10 (Childs Glacier), and remainder of cansk peer review edits, were discussed on the TM forum.


ovoss:
Problem with running datacheck.sh with branch as I named it. Will try again with simpler branch name.

PR #2723, opened 2019-04-09
ovoss-patch-999

ovoss:
Deletes ak.ak010.chi (Childs Glacier segment), peer review edits to some cansk files. All discussed on forum.


ovoss:
datacheck.sh appears to have completed successfully. Notes new unprocessed wpt file, which is probably ak.ak010.chi -- I'll delete that later.

Updates entry to follow separately.

PR #2727, opened 2019-04-10
datacheck of preview systems

si404:


michihdeu:
@si404 why do you have changed fps in my regions?
@ALL: Do we really need tons of VISIBLE_DISTANCE FPs?

si404:
@michihdeu I've not changed any of them. The order might have changed as I realphabetised the list to insert the new ones easier.

As for the visible distance errors - I'm decluttering the errors page, so one can actually find useful errors.

michihdeu:
Ok.

> As for the visible distance errors - I'm decluttering the errors page, so one can actually find useful errors.

That's great 👍 however, I'd prefer getting rid off them at all (or disable option in wpt editor and on datacheck.php, default = off)

PR #2748, opened 2019-04-16
datacheck

rmihailucian:


michihdeu:
@rmihailucian "DN2A/22" is not a good name for DN22 wp

rmihailucian:
Let it go, will correct tomorrow

michihdeu:
Sure, tomorrow is fine. When I see anything being wrong, I just wanna note it but definitely don't expect that it must be fixed "immediately".

PR #2754, opened 2019-04-17
Europe updates

si404:


yakra:
This update came to my attention because I noticed the number of hidden points marking the end of user travels had increased.
http://travelmapping.net/hb/index.php?&r=srb.e80&u=panda80
http://travelmapping.net/hb/index.php?r=srb.e80&u=bubaclex

>  2019-04-17;Serbia;E80;srb.e80;Route removed from R259 and onto A4 between north of Pirot and west of Bela Palanka

This threw me off until I inspected the old & new versions of the file.
Looking at **bubaclex**'s transition point, I initially interpreted the "west of Bela Palanka" point as the "north of Pirot" point.
A more sure way of pinning down the transition points of the realignment would be to refer to specific waypoints -- and visible ones.

IMO, old temporary transition points (such as freeway <-> surface road), in use as a route is upgraded should be left in place to allow users to most accurately mark their travels; some people may still need them, or need them in the future.
[My own travels on NB1](travelmapping.net/hb/index.php?u=yakra&r=nb.nb001) are good example; 8 of the 16 waypoints I use are at closed locations.

* IIUC, `R221_Tem` was a point on E80 back when it was on R259. Never been access on A4, I presume.
  - Rather than just keeping on mapping **panda80**'s travels incorrectly, with no error line appearing in panda80.log to clue him in to the need for a change, it may be better to simply have a standard shaping point here & break the .list.
  - Or alternately, with only panda80.list affected, `R221_Tem` could become an AltLabel for `R259_PirN`, which *I think* ought to map his travels correctly.
* Suggest also a closed point roughly corresponding to the old `Pir_W`.
* `BelaPal_E` looks like it dates to the transition off [this alignment](https://github.com/TravelMapping/HighwayData/blob/3a70c0a03198842ceb93a53e1b0523f5035b6997/hwy_data/SRB/eure/srb.e80.wpt), and should be unhidden, much like the historic transition points in Canada.
* Similar for other hidden points...

ping @rmihailucian; it looks like you maintain panda80.list.

rmihailucian:
@yakra Thank you for making me aware of this. I removed almost the whole E80 part from my list, as the motorway was not there when I traveled in the area, and drove on what's now R259, from the small those days existing A4 section near the border, till R221. Maybe someone will draft the R Roads sometimes, and I will be able to correctly map my travels there.

si404:
I've replaced + with * for the in use points (and matched them on both A4 and E80 despite them having different points in use).

PR #2759, opened 2019-04-19
I-380, PA 307, PA 347, PA 348, PA 390, PA 423, PA 435, PA 502, PA 507, PA 590, and PA 690 Recentering

Markkos1992:
New Routes: PA 390 TRUCK (Hawley) and PA 690 TRUCK (Moscow)

Other Changes:
I-81: Added 180A to line up with PA 502 and 184A to line up with PA 307. Also added +X15A and +X15B.
I-84: Added +X00 (to keep line south of PA 348)
I-380: See PA 435 and PA 507 for information regarding Exit 13.
US 6: Added Saddle Lake Rd (SadLakeRd) and +X15A.  Replaced +X4C with Lithia Valley Rd (LitValRd).  Replaced SprSt with Church St (ChuSt).  Replaced +X22 with Silk Mill Dr.  Replaced DeerRunRd with Shoshola Falls Rd (ShoFalRd).
US 6 BUS (Carbondale): Replaced PA 347 with Hull's Creek Rd (HulCreRd).  PA 347 was marked as closed and placed at the old connecting ramp location as shown on [Historic Aerials (2005)](https://www.historicaerials.com/viewer).  Added Greenfield Rd (GreRd), +X2, West Side Dr, and 8th Ave (8thAve_E).
US 11: Added McAlpine St (McASt), Rocky Glen Rd (RocGlenRd), Elm St, River St (RivSt), and Shady Lane Rd (ShaLaneRd).
PA 106: Added 42nd St.
PA 107: Added Washington Ave (WasAve).
PA 191: Added Forks Bridge Rd (ForBriRd), Golf Park Dr, Middle Creek Rd (MidCreRd) and Hoadlys Rd (HoaRd).
PA 196: Added Nevin Rd (NevRd).
PA 247: Added Hill St (HillSt_S), Grassy Island Ave (GraIslAve_W), Constitution Ave (ConAve_E and ConAve_W), Main St, and Green Grove Rd (GreGroRd).
PA 296: Added Cortez Rd (CorRd) and Owego Tpk (OweTpk).  Replaced RobRd with Jubinsky Rd (JubRd).
PA 307: Added Blue Shutters Rd (BluShuRd), +X1A, Harrison Ave (HarAve), Keyser Ave (KeyAve_W), Newton Ransom Blvd (NewRanBlvd), Falls Rd (FalRd), Dickinson Rd (DicRd), and Buttermilk Rd (ButRd).
[...]

PR #2762, opened 2019-04-20
US 6: Readded SprSt as Alternate Label for ChuSt

Markkos1992:
Put SprSt in as alternate label for ChuSt.

This was done while looking at the latest User Data list files and realizing that I broke griffith's with last night's Pull Request (https://github.com/TravelMapping/HighwayData/pull/2759).


Markkos1992:
After griffith's update, no one else is using SprSt so I will be removing it for real in a later pull request.

PR #2799, opened 2019-05-01
Poland and Catalan reviews

si404:


michihdeu:
@si404 GSJ is not a good wp name @C66

PR #2805, opened 2019-05-04
Europe stuff

si404:


michihdeu:
@si404 why Andalusian green routes to eursf**r**?

si404:
Because then they can be checked before added to the active system.

michihdeu:
Thanks. And I finally got what "r"-suffix means. Review 👍

PR #2820, opened 2019-05-07
ROU

michihdeu:


michihdeu:
@rmihailucian DJ707Lfol is unprocessed - missing in both csv files:
`hwy_data/ROU/roudj/rou.dj707lfol.wpt`
In addition, DJ707L misses city entry too.

PR #2828, opened 2019-05-09
turd corrections (D600s und

rmihailucian:


rmihailucian:
D600s and the last part of D700s

PR #2912, opened 2019-06-08
catch up while I can

si404:


michihdeu:
@si404 shouldn't _R.A. Bettolle-Perugia_ be called _Raccordo Autostradale Bettolle-Perugia_ as you've done for the other routes in itaa*.csv files?

PR #2965, opened 2019-07-15
IDF

michihdeu:


michihdeu:
@jteresco since there are no duplicate routes, I think there is no issue with graphs as we had when @si404 had tested the split. If I'm wrong, just comment the system.csv line.

jteresco:
Even if so, it's not a big problem as long as it's temporary.

PR #2973, opened 2019-07-17
NM and QC edits

ovoss:
Per forum discussion.


ovoss:
Unable to run datacheck.sh. Can't find the patch number I need to provide to the system.

jteresco:
The changes are straightforward enough that I'm fine to merge these in without a successful datacheck.

PR #2988, opened 2019-07-21
NV and QC changes

ovoss:


ovoss:
Truncation of QC A-19 per forum discussion. Also fix of NMP on NV 208's west end.

ovoss:
Datacheck successful.

This pull request was simple, but forthcoming cansk changes (still being worked out) will be difficult and definitely will need datacheck.sh

PR #3073, opened 2019-08-23
1D

neroute2:


neroute2:
> There is another interchange between +X1 and +X2 which must be added
> 
Has it been opened? In October 2017 it was barricaded.

> Some long understore errors (LS), see http://yakra.teresco.org/manual/manual_wayptlabels.html

Fixed. I chose town names based on what's signed in both directions on 1D.

> About your U turn name changes: `Use a _U suffix for interchanges that are nothing more than a U-turn ramp. If more than one is needed for the same highway, use _U1, _U2, etc. `

That's only https://www.google.com/maps/@32.5614268,-116.047661,3a,28.5y,170.13h,99.29t/data=!3m6!1e1!3m4!1sPTuUNfQ7oMiew--iC9voJA!2e0!7i13312!8i6656 , right? (though arguably this is simply part of the interchange to the southeast with a wide median...)
> 
> I'll go into details once these issues will be addressed for 1D and 2D.

I think I've fixed all the labels.

michihdeu:
> > PlaSal is not a possible end point of a travel since off- and on-ramp are on the southbound carriageway only and it's just a dead end. Please remove it again.
>
> I didn't realize that was a rule. Is this also true for highways with numbered exits?

I think the situation is similar to a rest area which never gets a wp except when there is an exit heading to other roads. I never had a "numbered" rest area. On the other hand, U turn on expressways always get wps.

> AreaDes is an interchange where you can turn around...

Yes, sorry. My bad.

> > CArt27Con has "four words" but the road name should have three words only. I'd leave Art27Con as-is.
>
> Should we be omitting the road type (Calle) if we know it?

I would not but since the system was untouched by TM... In CHM times (I was not yet on board back then), Tim was very strict with the rules. OSM also indicates it without Calle west of 1D.

> > PerLibSur is also a wp in use. I don't get why we should rename it to LibS. Sur should not be S and why to omit Per?
>
> I thought we abbreviated directions, but I guess not in names.

Exactly!

> But PerLibSur is short for "Periferico y/o Libramiento Sur", translated as "Beltway and/or South Bypass". In other words, someone didn't know how to correctly map alternate names.

yep. But do we need alternate names? We don't do it in Spain, Italy or France et cetera. I think that PerLibSur is fine.

> I couldn't find any signs

Me too.

michihdeu:
> Should RanDes be deleted?

No, OSM is wrong. Check GM or GS.

michihdeu:
I think you got the "alt label" thing wrong. There is always just ONE (correct) wp label. Alt Labels are only used when the wp label needs to be changed to avoid breaking user list files. It's an invisible alt name.

neroute2:
I don't get what you're saying. I added alt labels to keep in use labels in the list.

michihdeu:
Sorry, I've not seen that you've truncated the route meanwhile.

yakra:
> You have renamed AveEst to SanAnt. Av Estero (Google Maps) is a street name and San Antonio is a city name. Using a visible city name instead of a non-visible street name is according to the manual but we usually prefer street names meanwhile. @si404 @yakra please look at this and give a recommendation!

Hmm. Yes, street names have generally been preferred over city names, but be darned if the manual doesn't say *"Truncated, **visible** cross road name"*...

yakra:
What's the story with the truncation? Was that bit part of vanilla Mex1?

>> You have renamed AveEst to SanAnt. Av Estero (Google Maps) is a street name and San Antonio is a city name. Using a visible city name instead of a non-visible street name is according to the manual but we usually prefer street names meanwhile. @si404 @yakra please look at this and give a recommendation!
>
> Hmm. Yes, street names have generally been preferred over city names, but be darned if the manual doesn't say *"Truncated, **visible** cross road name"*...

Getting back to this for a bit, I'm not 100% sure what we're doing is the right thing.
What my ultimate recommendations would be here hinges on how we end up interpreting the manual...

To step backwards thru the manual's stipulations, and tinker with them a bit:

> 5. Truncated junction name - for freeways/expressways with official destination-based named junctions (e.g., some European countries use these).

What makes an *official* destination-based named junction? I'd think, a bit more than "something appearing on a sign"...

> Waypoints for inadequate highway data or other rare situations. Use these only when you are unable to properly label the waypoint according to the above options.

The implication here is that using named junctions would fall under the heading of "inadequate highway data or other rare situations". If this should be used only when "unable to properly label the waypoint according to the above options", then ... are we able to do that?

> 4. Truncated, visible cross road name. Magothy Bridge Road, etc.

The crux of this item, then, is what's a **visible** cross road name? Visible from *the road we're traveling on* specifically, such as on a BGS or exit sign? Or would that mean, visible in general; known to the traveling public? E.G., On a blade sign wherever...
To that end, what if a "visibly numbered cross road designation" is not visibly numbered on our BGS?

My knowledge of Europe is pretty limited, and I don't know what systems use named junctions, or how that works out. @si404, what's your experience here?

Whatever the case in the end, whether street names or junction names are used should be consistent; right now I see a mix of the two.
(And then, there'll be stuff like [this](https://www.google.com/maps/@32.4464093,-117.1040752,3a,63.5y,153.97h,90.79t/data=!3m9!1e1!3m7!1swskfVoT6xtfnU5ZeLOzDwg!2e0!7i13312!8i6656!9m2!1b1!2i40) gumming up the works...)

---

**CArt27**: relabel? What's the genesis of this name format?
If we *do* end up going with destination-based named junctions, then maybe simply `Ros`.
**CCliRuiz**: Signed for Rosarito Sur.
Ah, OK, I missed **CEns** between them. Looks like we're using 'C' as the standard abbreviation for "Calle", then? If so, and if we retain street-name labels here, these labels are good.

**MEX1_Nue** -> MEX1_Pue
It's signed for MEX1D, but ISTR there being a comment, either on GitHub or the forum, about designations often being "off" like this? An extra 'D' where there shouldn't be one; a missing 'D' where there should be?

**MEX1_Mis**:
This exit only serves the NB lanes, maybe move S halfway to the RIRO for the SB lanes?
Alistos is only signed NB; La Mision is signed both NB and SB. In this context, michih's preference for `MEX1_Mis` makes sense.

**PuePie** -> PunPie

Revert **Sali** -> Sal, truncating to 3 letters.
`Sals` for Salsipuedes is sufficient to disambiguate; we only want to expand one of these labels out to 4 letters.

michihdeu:
> My knowledge of Europe is pretty limited, and I don't know what systems use named junctions, or how that works out. @si404, what's your experience here?

Why Europe? Mexico is Mexico. Should we go strict with the [rules](http://travelmapping.net/devel/manual/wayptlabels.php)?

Europe is according to the rules and their order but 
`4. Truncated, _visible_ cross road name`
is interpreted in a different way. "visible" just means indicated on maps (OSM or GM) but not visible in the field (GSV).
 
I'm not sure about the meaning of 5.:
`5. Truncated junction name - for freeways/expressways with official destination-based named junctions (e.g., some European countries use these).` What's an "official destinated-based named junction"?

I think we should discuss these details on the forum with the goal to find a common way (maybe region-specific exceptions) and **revise the manual.**

si404:
Is a destination-based junction name like "Modena Nord" or "Firenze Scandicci" on the Italian A1? These would be opposed to location-based junction names like "Hook Moor" (OK, that's M1 J48 or A1(M) J43), or "Air Balloon" (A417/A436 junction named after the pub there) seen in Britain.

I should note that while "Europe is Europe" we've followed the US-centric manual guidelines on waypoint labels.

michihdeu:
Ah, I see, junction names!

http://travelmapping.net/hb/index.php?r=deubw.b029

`BruSt` would be `RemW` (Remshalden-West)
`LorStr` would be `SchGmuW`
`Her` would be `MogE` (Mögglingen-East)

I like it somehow but I never followed the rule.

neroute2:
> What's the story with the truncation? Was that bit part of vanilla Mex1?

It's locally maintained and (ignoring implied TOs) not signed as 1D. I've created a future file for mexsf: https://github.com/neroute2/HighwayData/blob/mexsf/hwy_data/MEX-BC/mexsf/carresc.wpt
> 
> > Waypoints for inadequate highway data or other rare situations. Use these only when you are unable to properly label the waypoint according to the above options.
> 
> The implication here is that using named junctions would fall under the heading of "inadequate highway data or other rare situations". If this should be used only when "unable to properly label the waypoint according to the above options", then ... are we able to do that?
> 
> > 1. Truncated, visible cross road name. Magothy Bridge Road, etc.
> 
> The crux of this item, then, is what's a **visible** cross road name? Visible from _the road we're traveling on_ specifically, such as on a BGS or exit sign? Or would that mean, visible in general; known to the traveling public? E.G., On a blade sign wherever...
> To that end, what if a "visibly numbered cross road designation" is not visibly numbered on our BGS?

I've looked for street signs in the area of the interchange. Only in those cases where I cannot find any at all have I used the place name that signs use. I don't trust OSM or the Goog when it comes to Mexican street names (or U.S. street names, for that matter).
> 
> Whatever the case in the end, whether street names or junction names are used should be consistent; right now I see a mix of the two.

We can't be consistent with street names because we don't always have a street name (even official GIS data says S/N, sin nombre).

> (And then, there'll be stuff like [this](https://www.google.com/maps/@32.4464093,-117.1040752,3a,63.5y,153.97h,90.79t/data=!3m9!1e1!3m7!1swskfVoT6xtfnU5ZeLOzDwg!2e0!7i13312!8i6656!9m2!1b1!2i40) gumming up the works...)

As with a street that changes names at the intersection, we choose one. Probably the one that is prominently signed in both directions.
> 
> **CArt27**: relabel? What's the genesis of this name format?
> If we _do_ end up going with destination-based named junctions, then maybe simply `Ros`.
> **CCliRuiz**: Signed for Rosarito Sur.
> Ah, OK, I missed **CEns** between them. Looks like we're using 'C' as the standard abbreviation for "Calle", then? If so, and if we retain street-name labels here, these labels are good.

Yes, C is the abbreviation for Calle: https://www.waze.com/forum/viewtopic.php?p=105535 (also seen on signs)
> 
> **MEX1_Nue** -> MEX1_Pue

I was thinking Nuevo is a more unique name than Puerto (Port). We could easily have two towns named Puerto x. But there's no such confusion here, so changed.

> It's signed for MEX1D, but ISTR there being a comment, either on GitHub or the forum, about designations often being "off" like this? An extra 'D' where there shouldn't be one; a missing 'D' where there should be?

Yes, and even cases where they don't use the correct number at all: https://www.google.com/maps/@19.4377585,-99.2971274,3a,46.8y,19.75h,95.3t/data=!3m6!1e1!3m4!1skPGRAQtrdHKsCRxNJv1uSA!2e0!7i13312!8i6656 (to the right should be 134)
> 
> **MEX1_Mis**:
> This exit only serves the NB lanes, maybe move S halfway to the RIRO for the SB lanes?

Eh, I'd think it should be at the bridge, no matter how the ramps are arranged.

> Alistos is only signed NB; La Mision is signed both NB and SB. In this context, michih's preference for `MEX1_Mis` makes sense.

That's also my preference, which is why I changed it. The current file has Ali.
> 
> **PuePie** -> PunPie

Fixed.
> 
> Revert **Sali** -> Sal, truncating to 3 letters.
> `Sals` for Salsipuedes is sufficient to disambiguate; we only want to expand one of these labels out to 4 letters.

I wasn't aware there was a standard for these. If we had inventoried the route in the other direction, would we use Sal and Sali?

michihdeu:
> > Revert Sali -> Sal, truncating to 3 letters.
> > Sals for Salsipuedes is sufficient to disambiguate; we only want to expand one of these labels out to 4 letters.
>
> I wasn't aware there was a standard for these. If we had inventoried the route in the other direction, would we use Sal and Sali?

We always truncate to 3 letters except when the word does acually just have 4 letters at maximum.

http://travelmapping.net/devel/manual/wayptlabels.php

Another exception:

> If two named cross roads have different names but would have identical labels (Milford Parkway & Millville Parkway both would be MilPkwy), either add a 4th letter to a 3-letter part of one of the labels, or choose 3 different letters for that label.

PR #3074, opened 2019-08-23
2D

neroute2:


jteresco:
Just requesting some review by experienced contributors on these.

neroute2:
> Why have you reversed the route order? This way Github indicates that all wps have been changed. Hard to see your modifications.

That's the way the km posts go, and the way exit numbers will go if Mexico ever starts using them large-scale.
> 
> When working on a system like this, you should also check the positions of the wps. For instance, 2D +X12 and +X6B are far off
> 
> You have added LagSal and CanDeGua wps but should also remove +X10B and +X10A now since the route is not exceeding the red lines in wpt editor. The red lines indicate the "tolerance". A situation with a very wide median like between MEX2D_U2 and MEX2_Rum is the only exception.
> 
> What's the reason for adding these wps? It's fine to replace hidden wps by visible wps - some hwy data manager do it this way, I like is less and would not exagerate - but I'd just like to know the reason. Do you need them for your own travels?

Since this is part of the "freeway" network, I added anything that has a green guide sign.
> 
> I'd call Canon de Guadalupe wp just CanGua but that's just my way. CanDeGua might be fine too but just do it consequently all over Mexico. It's not possible for wp suffixes, see LS errors. In addition, AaaBbbCcc is fine but AaaBbbCccDdd isn't.
> 
> You've also renamed MEX2D_U2 to ParLaCue (Parador de Cuesta) which is fine to me. @yakra, @si404, what do you think about this?
> 
> Ban -> FraBan (Fracc Banchetti) is fine but Ret (Retorno) might be even better since the wp label is shorter.

Retorno means U-turn, not a good name for a point. Fracc means neighborhood and probably isn't necessary.
> 
> The wp table in the wpt editor indicate data errors. The route had 4 visible distance errors (VD) and 1 sharp angle error (SA) which are marked als false positives (FPs). See tooltip.
> Now, it has 2 VD, 1 SA, 2 more then one underscore (__) errors and 5 long underscore (LS) errors. That means, renaming the wps was not according to the rules.

If I don't have a copy of the rules, how can I follow them?
> 
> You've split MEX2_TecE into 2 wps. The positions of the new wps look fine but I'm not sure that we should really split the wp. Since you have no valie wp name (MEX2_W_TecE + MEX2_E_TecE), I'd leave as-was.

It's clearly two separate interchanges with an overlap...
> 
> PlaDart is fine :)
> 
> Renaming Blvd2000 to BC201 is also fine since it's not just indicated on OSM but really signed according to GSV.

I couldn't find any BC 201 signs, so I changed it to Blvd2000 (this is how it's signed on 1D).
> 
> BlvdLicHecTerTer might be called BlvdLHTT but I like the original one "BlvrHecTer" with just renaming Boulevard abb. -> BlvdHecTer.
> 
> http://travelmapping.net/logs/pointsinuse.log indicatess that wps in use are: BLVRLAZCAR MEX2_MEXE MEX2_RUM. You should NOT rename these wps when it is an active system (it is) since you would break user list file entries. You need to keep them als alternative label, for instance:
> 
> BlvrHecTer http://www.openstreetmap.org/?lat=32.539315&lon=-116.904917
> ->
> BlvdHecTer +BlvrHecTer http://www.openstreetmap.org/?lat=32.539315&lon=-116.904917

I couldn't figure out how to determine if waypoints are in use. No manual, remember?
> 
> Why have you've truncated the route at west end? This should be discussed on the forum first since you are new. If we agree on truncating, we need an update entry.

It's pretty clearly just a major city street there.

michihdeu:
> That's the way the km posts go, and the way exit numbers will go if Mexico ever starts using them large-scale.

That's fine. Good catch!

> I changed it back for now.

Not necessary. It was just about Github. It indicates changes in green and red but when you change the order, all lines are red and I cannot directly see what you've changed. When changing it now - I know, you already did - I just have two "red lists".

> Retorno means U-turn, not a good name for a point. Fracc means neighborhood and probably isn't necessary.

Sorry, I didn't know that it's just for the U turn. So, Ban should stay as is.

> It's clearly two separate interchanges with an overlap...

It's not that clear to be. I'd like to get other opinions.

> I couldn't find any BC 201 signs, so I changed it to Blvd2000 (this is how it's signed on 1D).

I [could](https://www.google.com/maps/@32.5448038,-116.8527942,3a,15y,259.41h,116.63t/data=!3m6!1e1!3m4!1sC_gZJCaJVjSdVc2y_Ymj_A!2e0!7i13312!8i6656).

> I couldn't figure out how to determine if waypoints are in use. 

Sure, that's why we make the review.
1. http://travelmapping.net/devel/tools.php
2. [Waypoint labels in use by current TM users](http://travelmapping.net/logs/pointsinuse.log)
3. Use the find option of your browser and enter the wpt file name without file extension: `mexbc.mex002d`
4. Copy the wps in use: `BLVRLAZCAR MEX2_MEXE MEX2_RUM`
5. Paste it to the small input field in wpt editor
6. Press "Load In-use Labels"
7. The wpts in use are in red on the table now

> It's pretty clearly just a major city street there.

However, it was drafted like this. ls the segment to the west not signed? Where is the last sign. Maybe it should even be truncated to BC201 as GM suggests?

neroute2:
> > I couldn't find any BC 201 signs, so I changed it to Blvd2000 (this is how it's signed on 1D).
> 
> I [could](https://www.google.com/maps/@32.5448038,-116.8527942,3a,15y,259.41h,116.63t/data=!3m6!1e1!3m4!1sC_gZJCaJVjSdVc2y_Ymj_A!2e0!7i13312!8i6656).

Good catch. And strange shield.

> > It's pretty clearly just a major city street there.
> 
> However, it was drafted like this. ls the segment to the west not signed? Where is the last sign. Maybe it should even be truncated to BC201 as GM suggests?

Mexico is weird with signage in cities (lots of implied 'to' on signs), and sources disagree here. Capufe (toll) maintenance ends at BC 201, but GIS data shows the route west to BlvdLicHecTerTer is under federal control. I could go either way.

BTW 1D is very similar; toll maintenance actually ends at the beach exit where it curves along the border, with the remainder of the freeway inventoried as a branch of 1.

michihdeu:
> Mexico is weird with signage in cities (lots of implied 'to' on signs), and sources disagree here. Capufe (toll) maintenance ends at BC 201, but GIS data shows the route west to BlvdLicHecTerTer is under federal control. I could go either way.

That's exactly why I suggest discussing it on the forum first. I think that there are two options: Leave as-is or truncate it to BC201. Your solution with truncating it half-way is not an option for me.

neroute2:
http://forum.travelmapping.net/index.php?topic=3167.0

yakra:
I still haven't gotten the chance to pay as much attention to this as I would like.
I see that the latest commit brings a number of point labels in line with the manual; that's good. Still wanna check out the rest in more detail.

yakra:
Just a quick response to the bits where @michihdeu pinged me. I've still only mostly skimmed this thread... :(

> You've also renamed MEX2D_U2 to ParLaCue (Parador de Cuesta) which is fine to me. @yakra, @si404, what do you think about this?

ParLaCue looks a bit questionable as a name; whereas `MEX2D_U2` is an officially sanctioned type per the [manual](http://travelmapping.net/devel/manual/wayptlabels.php).
I'd say that qualifies as a kind of "Visibly numbered cross road designation" (type 3), rather than a truncated junction or town name (type 5/7/8).



>>> You've split MEX2_TecE into 2 wps. The positions of the new wps look fine but I'm not sure that we should really split the wp. Since you have no valie wp name (MEX2_W_TecE + MEX2_E_TecE), I'd leave as-was.

>> It's clearly two separate interchanges with an overlap...

> It's not that clear to me. I'd like to get other opinions. @yakra @si404

I could go either way on this one. There is a however-short multiplex of 2 & 2D involved, so there's that. I'm OK with splitting it.

PR #3085, opened 2019-08-28
US66 Historic in New Mexico

si404:
ping @oscar


yakra:
It appears that old 66 is on the frontage road, not I-25 itself.

jteresco:
Will wait for the word that this is good to go before merging in.

ovoss:
If so, should the historic route be using I-25(xx) waypoints?

si404:
I've changed the I-25(xx) waypoint labels on the points on the frontage roads.

si404:
As I've fixed the issue @ovoss raised, I'm going to merge this. Should there be any further issues, they can be dealt with in a further pull.

PR #3098, opened 2019-09-02
TUR O5 new openings

rmihailucian:


michihdeu:
On the other hand, TUR O5 has no traveler. That means, the `altRouteNames` are not required.

rmihailucian:
> On the other hand, TUR O5 has no traveler. That means, the `altRouteNames` are not required.

I have just added them. Should I keep the alt Label or remove them?

michihdeu:
Leave them but remove the blank

rmihailucian:
Improved updates.csv, adding the exit numbers of the new routes.

michihdeu:
Data check successful!

PR #3109, opened 2019-09-06
CA 16 and 14U changes

ovoss:
Splits preview route CA 16, truncates 14U, Also minor changes to other CA routes. Discussed in forum.

@si404 New 14U endpoints added to US6HisSan route file, to keep two routes in synch.


ovoss:
Datacheck successful.

PR #3115, opened 2019-09-08
extended 5D and added branches

neroute2:
Note that there are a lot of unrelated creations and deletions in this branch, since I didn't realize that saving to a new branch would include everything in my mexsf branch. Only five files are actually created or changed, all relating to 5D.


jteresco:
Thanks.  Hoping a couple of our experienced highway people can take a look to make sure this is good to go.

yakra:
**mexed_con.csv:**
The `_con.csv` files are used to connect single routes that have been chopped at regional boundaries. For example:
https://github.com/TravelMapping/HighwayData/blob/02b9df967e4a6e30c6443d008cafb22a5545408b/hwy_data/_systems/usame_con.csv#L69
Note the two NH segments of ME113.
What we have here are 3 separate EMEX5D routes, all in MEX-EMEX.
We'd want 3 separate CSV lines, each with just one root.

I haven't looked at the contents of the files yet, or the changes to mexemex.emex005d.wpt.

neroute2:
Ah, I think I fixed mexed_con.

PR #3146, opened 2019-09-15
NZL+NSW+SVK+SCT

si404:


michihdeu:
@si404 Did you forget running data check?

`ABORTING due to 1 errors:
1: Route contains fewer than 2 points: ausnsw.m001uru (0 total points)`

You deleted the route but didn't removed it from csv files.

@jteresco Can I access the output file anyhow? I thought it's `/home/www/tm/logs/siteupdate.log` on noreaster but it isn't. It's identical to http://travelmapping.net/logs/siteupdate.log - the previous one from the last successful run.

jteresco:
Sorry, I launched the update and never looked again at it before bed.  The log doesn't go in place when the update fails.  I'll make a quick fix and re-run.

PR #3159, opened 2019-09-20
Prep for cansk activation

ovoss:
Adds SK1, SK16, SK16A, and SK16B, which are copies of the corresponding Trans-Canada Highway route files. Also tweaks waypoint labels for QC109.


ovoss:
Datacheck successful.

PR #3160, opened 2019-09-21
update per forum

the-spui-ninja:
Let me know if I need to edit updates.csv; I don't think I do


jteresco:
None of these look newsworthy to me, so I don't think any updates entries are needed.

yakra:
> Let me know if I need to edit updates.csv; I don't think I do

As it stands now, the existing point has just been moved; there's nothing to change in a .list file, nothing to de-clinch. So in this context, no need to add an updates entry.

I wonder though, is it worth it to add in points for the old alignments (and thus push this into newsworthy territory)?

**ND58:**
```
...
39thSt http://www.openstreetmap.org/?lat=47.978259&lon=-104.023232
*OldND58 http://www.openstreetmap.org/?lat=47.992650&lon=-104.006533
ND1804 http://www.openstreetmap.org/?lat=47.997072&lon=-104.002410
```

**ND1804:**
```
...
153rdAve http://www.openstreetmap.org/?lat=47.998163&lon=-103.996335
ND58 http://www.openstreetmap.org/?lat=47.997072&lon=-104.002410
*OldND58 http://www.openstreetmap.org/?lat=47.999083&lon=-104.006603
MT/ND http://www.openstreetmap.org/?lat=48.002529&lon=-104.044261
```
...with `153rdAve` being relocated here to serve double duty as the break from the old alignment.

Compare where [US85BusWat](http://travelmapping.net/hb/?u=harvey&r=nd.us085buswat&lat=47.784227&lon=-103.319764&zoom=13) meets [its parent](http://travelmapping.net/hb/?u=harvey&r=nd.us085&lat=47.784227&lon=-103.319764&zoom=13).
There, we have the option to de-clinch...
0.81 mi of US85 & 0.53 mi of US85Bus @ the S end, and
0.64 mi of US85 & 0.28 mi of US85Bus @ the N end.
Here, we'd have the option to de-clinch...
0.37 mi of ND58 & 0.54 mi of ND1804.
@ovoss, your thoughts? How's this compare to "the usual amount of stranded pavement"?

FWIW, if the option were available, I'd declinch these segments in [a list file I'm maintaining based on my parents' travels](http://travelmapping.net/user/mapview.php?units=miles&u=harvey&rg=ND).

PR #3162, opened 2019-09-21
Fix hidden junction on A-85

ovoss:
Problem seems to be extra hidden point on concurrent TCH (Main), which this pull removes.

### DIFF
--- a/hwy_data/QC/cantch/qc.tchmai.wpt
+++ b/hwy_data/QC/cantch/qc.tchmai.wpt
@@ -179,7 +179,6 @@ A-85(24) http://www.openstreetmap.org/?lat=47.591404&lon=-68.750274
 +X556793 http://www.openstreetmap.org/?lat=47.583821&lon=-68.714504
 *OldQC185_A http://www.openstreetmap.org/?lat=47.575339&lon=-68.708764
 +X116062 http://www.openstreetmap.org/?lat=47.556024&lon=-68.688240
-+AvePri http://www.openstreetmap.org/?lat=47.547668&lon=-68.658081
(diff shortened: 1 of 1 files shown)


yakra:
Oops, I meant to merge in #3163, but went to the wrong browser tab.
I'll take a closer look around and see if this was the right thing to do.

yakra:
Hm, oh bother.
There are ramps there unless things have changed since [Sep 2018](https://www.google.com/maps/@47.5467457,-68.6549844,3a,15y,298.19h,95.33t/data=!3m9!1e1!3m7!1sm1qLMqirQ-JlJKTCapp5Ag!2e0!7i13312!8i6656!9m2!1b1!2i40), 0.6 mi from the existing exit 14 point. This should arguably be included (`4A`?) in both files...

jteresco:
Thinking about an afternoon site update to get cansk active on site.  Should I wait for a further change on this one before starting it?

yakra:
Not IMO. I was going to write a comment on #3160, but that doesn't really need to hold us back either.

yakra:
lab2 update in progress.

ovoss:
Nothing has changed, AFAIK, since I decided to treat the Ave. Prinicipale ramp as part of the exit 14 (QC 295) interchange. I originally kept that in the TCH file as a hidden point only because someone was using it. It's no longer in use.


I don't have time to deal with this right now. In any case, it shouldn't hold up the hidden junction fix, or anything else.

yakra:
Hidden junction fix is in.
If you want to treat this as one interchange & thus one point, fine by me.

PR #3218, opened 2019-10-11
Add I-5 truck route to usai

ovoss:
Adds I-5 truck route in Newhall CA to usai, per forum discussion.

Also, updates SK 7 for newly-opened bypass of Vanscoy.

Updates table entries for both to be submitted in separate pull request.

Also, several California route files updated, in first batch of usaca peer review changes.


ovoss:
Datacheck failed due to csv file error. Will resubmit with corrections.

PR #3222, opened 2019-10-12
FRA, POL, BGR openings

si404:


michihdeu:
@si404 bgr A2 update entry is incorrect. The exit is called I3_E instead of I3_N

PR #3273, opened 2019-11-03
cansk.csv error fix

ovoss:
Last-minute fix to Alt Route Name didn't make it into pull request #3271. This only moves the AltRouteName field for deleted sk.sk011reg to the right line, and corrects the field format.


jteresco:
This had a merge conflict, but I believe I resolved it properly based on @ovoss comment above.

PR #3306, opened 2019-11-13
creating two devel systems: mexsf and mexdfeje

neroute2:
Hopefully I'm doing everything correctly here.


jteresco:
I'll give it a shot in tonight's update.  If the update fails from anything here, I'll see if I can fix up, and if not, I'll comment out the systems.csv entries until it can be taken care of.

jteresco:
Successful site update.  The new systems are in at the devel level.

neroute2:
Thanks :)

PR #3316, opened 2019-11-14
WY cleanup: Johnson County + Bighorn NF

the-spui-ninja:
I didn't include this: http://forum.travelmapping.net/index.php?topic=3327.0 
just to give more people time to comment on it.


jteresco:
The wy.us016.wpt update broke user list files, as it introduced a second I-90BL_W label.  I'm removing that new label and restoring the old I-90BL_N for now.

the-spui-ninja:
Sorry about that, it slipped through the cracks. I think #3323 should fix that problem.

PR #3317, opened 2019-11-15
GES/IDF

michihdeu:


rickmastfan67:
@michihdeu Might want to double check your 'update' entry for the update page.  You listed 'A16' for both of them in the 'route' part, instead of one each for N1 & A16.  Might want to correct that. ;)

michihdeu:
@rickmastfan67 Thanks 😄

PR #3368, opened 2019-11-30
TN: Delete TN 52 Truck (Celina)

mapcat:


yakra:
Updates entry for the (admittedly minor) reroute to TN52 proper too?

mapcat:
Fixed in https://github.com/TravelMapping/HighwayData/pull/3370

ovoss:
@mapcat, it looks like TN 52 could use points in Celina at the Dow Ave./Main St. and Brown St./Lake Ave. intersections, which appear to be where the new and old routings intersect.

If TN 52 were a brand-new route, we wouldn't need these points just one block away from the new DowAve_E point. But since the route changed since I and perhaps other TM users drove TN 52 through Celina, those points would let us modify our list files to reflect the new two-block gap in our TN 52 travels.

Oscar

mapcat:
Yeah, somehow I forgot that people other than myself have travelled that route...

https://github.com/TravelMapping/HighwayData/pull/3376

PR #3373, opened 2019-12-01
Patch 217

rmihailucian:
@si404 I added 2 WPs that I need on national roads (on N6 in Astorga at intersection with LE133 and on  NILer in Lerma, at the intersection with Calle de la Estacion).
@jteresco please wait on this one, until @si404 confirms it's ok


si404:
fine with me - merged

PR #3377, opened 2019-12-02
Fixing PA Alternate Route Names

Markkos1992:
http://forum.travelmapping.net/index.php?topic=3372.0

These three are currently not in-use by anyone based on searching Github.


yakra:
` usapa;PA;PA252;AltTrk;New;Newtown Square;pa.pa252alttrknew; ` still has a trailing space. May be best to delete this just to be safe.

Markkos1992:
It looks like I got it now.

PR #3394, opened 2019-12-12
TN: TN 109 realignment and new I-65 exit 121

mapcat:


rickmastfan67:
I think that the 'TN 109' mention in the US-31W update shouldn't have a space in it since you're specifically mentioning the point label being moved.

mapcat:
True, that was unintentional. @jteresco can you fix it?

jteresco:
Done.

mapcat:
Thanks!

PR #3407, opened 2019-12-16
CO fixes from forum and WY review for part of Fremont County

the-spui-ninja:
http://forum.travelmapping.net/index.php?topic=3346.0
http://forum.travelmapping.net/index.php?topic=3406.0

Fittingly submitted in Denver International Airport


jteresco:
I expanded the updates entries for the US 85 and US 87 relocations to include the exit numbers.  I know I need to update my .list to remove that segment from those two routes, since I've never taken the BL through Walsenburg.

the-spui-ninja:
Ah, thanks, didn't even think of that.

PR #3413, opened 2019-12-17
Update mar.n016.wpt

rmihailucian:


michihdeu:
@rmihailucian I think it should be read: `_MDi` Always 3-digits!

PR #3414, opened 2019-12-17
OCC, SCT, MYS, POL

si404:


michihdeu:
1: Duplicate root in con_route lists: mys.e032
4: route mys.e032sit not matched by any connected route root.

michihdeu:
@si404 I found it and will fix it now!

PR #3426, opened 2019-12-19
usatxre -> active

yakra:


jteresco:
Since two activations are happening on the same day, I had to resolve a conflict by keeping both new entries.  Simple enough, but there's now an extra commit into the LoneStar branch, I think.

PR #3448, opened 2019-12-22
roudj corrections

rmihailucian:
Corrections after the review from @michihdeu


michihdeu:
Data check successful

PR #3450, opened 2019-12-23
GA fix

michihdeu:


michihdeu:
@jteresco @ntallyn 

I've renamed the file which caused that last night's update failed.
It's possible that the file name was correct and the csv wrong. If so, change it again.

ntallyn:
> 
> 
> @jteresco @ntallyn
> 
> I've renamed the file which caused that last night's update failed. My local data check was successful now.
> It's possible that the file name was correct and the csv wrong. If so, change it again.

Your fix is correct. Thanks!

PR #3452, opened 2019-12-23
ROU A1 opening

rmihailucian:
A1 is being extended, Lugoj-Deva L3 is finally open.


michihdeu:
Data check successful!

PR #3480, opened 2019-12-31
Suceava bypass added

rmihailucian:


michihdeu:
@rmihailucian I'll add the missing csv entries.

michihdeu:
Damn, no. I can't. You've just extended the route. It's not a New route as described in updates.csv. I'll modify the entry....

rmihailucian:
Added it as a new route because I forgot to add it in the updates.csv when the first stage of the road was opened.

michihdeu:
My bad, sorry. Undone.

PR #3495, opened 2020-01-02
AR peer review 101-200 + OH us020hisnor point renaming

mapcat:


mapcat:
@si404 note new name of west endpoint for OH US20HisNor (from GMSV)

PR #3661, opened 2020-03-14
GA route updates

ntallyn:
HDX fixes for remaining GA non-bannered routes. Re-added GA40SprStM.


yakra:
ga.ga040.wpt: ChaSmiHwy -> GA40Spr

PR #3814, opened 2020-04-28
DEU/GES/ROU/TUR

michihdeu:


michihdeu:
@rmihailucian I've removed old FP entries from ROU and TUR.

btw: I think `tur.o007dem;O7;;;LABEL_SELFREF;` should be added!

PR #3879, opened 2020-05-17
start out naca (Central American routes)

neroute2:


michihdeu:
@neroute2 sorry, but CAM is not a valid region name, see [regions.csv](https://github.com/TravelMapping/HighwayData/blob/master/regions.csv). This pull request cannot be merged in @jteresco

jteresco:
More needed than changing the CAM to NA?

neroute2:
Given that cism and cisa exist, camca (or caca?) should be an appropriate system name. So yes, I will change CAM to NA in systems.csv.

michihdeu:
> More needed than changing the CAM to NA?

I've never drafted an international system. @si404 should know best how to deal with it.

si404:
camca;NA;Rutas Centroamericanas;red;3;devel

this is fine, and fits the pattern of other systems.

yakra:
While the system name doesn't *need* to start with a region code, I would still consider it best practice

jteresco:
> While the system name doesn't _need_ to start with a region code, I would still consider it best practice

nacar for "Central America Routes" in the North America continent?

neroute2:
> 
> 
> > While the system name doesn't _need_ to start with a region code, I would still consider it best practice
> 
> nacar for "Central America Routes" in the North America continent?

Most systems drop the 'r' for route. I'm not sure why na is better than ca, since the routes are only in Central America.

michihdeu:
>  I'm not sure why na is better than ca, since the routes are only in Central America.

Because there is no Central America! https://github.com/TravelMapping/HighwayData/blob/master/continents.csv

And that's correct: https://en.wikipedia.org/wiki/Continent#Area_and_population
We also don't have Middle East and Far East but just Asia.

neroute2:
Whatever. I don't feel like arguing so I'll change it to naca.

michihdeu:
> Whatever. I don't feel like arguing so I'll change it to naca.

👍

PR #3899, opened 2020-05-20
Split QC125

ovoss:
This is subject to comments -- @yakra, @si404, @vdeane, your thoughts? Jim, don't rush to merge this into the master.

Datacheck successful.


ovoss:
@vdeane posted on the forum that she's OK with the route split. @yakra and @si404 not yet heard from.

Unless they have a problem with the idea of a route split, I suggest we go ahead with this pull request. Any comments on details like waypoint labeling can be addressed later.

jteresco:
It's preview, so it's fine to break some lists.

PR #3906, opened 2020-05-22
New WPs in BY

rmihailucian:
@jteresco please hold on this one until @michihdeu confirms it's ok.
@michihdeu I added a few WPs that I need after my trip to Rothenburg today. Please check if it's ok.


michihdeu:
yep, fine!

PR #3913, opened 2020-05-25
AUT/BEL/CZE/DNK/HDF/HUN/LUX/NI/NLD/NOR/PRT/ROU/SVN/TUR

michihdeu:


michihdeu:
@rmihailucian I also cleaned-up PRT, ROU and TUR!

PR #3924, opened 2020-05-29
Update systemupdates.csv

ovoss:
canqc activation


michihdeu:
@ovoss The preview entry should be kept!

michihdeu:
btw: The system was exactly 3 years in preview

michihdeu:
@ovoss I've added it again!

PR #3925, opened 2020-05-29
Update systems.csv for canqc activation

ovoss:
This pull request, and related pull requests 3923 and 3924, activates canqc.


jteresco:
This is worthy of an extra site update this morning.  Thanks for all the work on this system!

PR #3943, opened 2020-06-03
datacheck / connected routes fixes

si404:


michihdeu:
@si404 please improve the updates.csv info!

I was not able to reconstruct my broken A8 travel without loading the old A8 wpt file into wpt editor.
Please indicate that `BI10` and `A8Gal` are the former `A8` route

PR #3956, opened 2020-06-08
BB/HE/HUN/LTU/LVA/MV/ROU/RP

michihdeu:


michihdeu:
@si404

yakra:
LOL, beecde4 rectifies a longstanding annoyance of mine, how one person just went and unilaterally changed the headers to all the .csv files to match DB fields. (Why not change to match, say, the variables used to store the info in the siteupdate program?) IMO, the headers should be for the benefit of the people who work with the CSVs the most -- the HighwayData managers.
Sure, I prefer the shorter CHM-legacy standard of "Abbrev" and "Root", but this is good!
:+1:

michihdeu:
@yakra I was confused when writing https://github.com/TravelMapping/DataProcessing/issues/338
The manual said `name` but the csv files had `groupName`
_We could also change the manual though._

yakra:
> @yakra I was confused when writing https://github.com/TravelMapping/DataProcessing/issues/338
> The manual said `name` but the csv files had `groupName`

Even the manual itself (I'm assuming that what's still listed in http://travelmapping.net/devel_new/manual/syshwylist.php#chopped) is what's inherited from CHM) isn't consistent with the headers used in the files in the CHM days.

> _We could also change the manual though._

*My* opinion? I like the brevity and *(long-since trampled-on :( )* familiarity of `System;Region;Route;Banner;Abbrev;City;Root;AltRouteNames`. It's still how I refer to these columns in my own head. While I like the idea of changing the manual to match **that**, it's... kinda less urgent now that very few CSVs actually say that anymore. :P

The manual already says `In processing, the first row is always ignored.`, which is good.
Maybe add something else to emphasize that you can call the fields what you want, whatever helps you remember the columns and keep them organized; just don't put any actual data there?
While we do "Use the header row to label the columns," we don't *have to* label them anything specific.
I kinda like this example:
https://github.com/TravelMapping/HighwayData/blob/0111c8e94b1a7ca5c367e4b2758428e7c2bfd7be/hwy_data/_systems/usawi.csv#L1

I think the headers as shown in the manual now do a good job of being descriptive.

PR #3979, opened 2020-06-16
Various MD tweaks

Duke87ofST:


ovoss:
Shouldn't MD5_S be renamed (perhaps to ToMD5_S), and moved west to the roundabout shown in ESRI WorldStreetMap? The way it is now, MD 373 is shown with two intersections with MD 5, but actually has only one (MD5_N, at Spine Rd.). My suggestion would show MD 373 running parallel to MD 5 between the roundabout and Spine Rd., especially if you keep the shaping point just west of MD5_N.

jteresco:
Will hold off on this one for tonight, pending comment or action on the suggestions by @yakra and @ovoss .

Duke87ofST:
I left the MD5_S point in the median of MD 5 because putting it in the middle of the new "roundabout" (it's not actually a roundabout, ESRI is wrong on that) would mean no graph connection here. Maybe it doesn't matter since there is one just to the north anyway, but that was my reasoning. I can shift it if we think that's better.
Also, per GMSV, the frontage road on the west side is signed as Accokeek Road.

Regarding the directional suffixes, shouldn't those depend on the direction of the intersecting route? MD5 is a N-S route, so why would there be MD5_E and MD5_W labels for it?

ovoss:
I think at some point MD5_S should be moved west. That way, there's only one graph connection with MD 5 rather than two, with the second a false connection since MD 373 no longer intersects MD 5 in the vicinity of MD5_S.

No other comments from me.

yakra:
I agree with @ovoss re moving the point to the roundabout. Graph connections aren't always appropriate when there are more accurate ways to represent a highway without.

Directional suffixes would depend on the direction of the intersecting route when multiplexes are involved. Since we have 2 separate non-multiplex junctions, we go to [the rule about](https://travelmapping.net/devel/manual/wayptlabels.php#underscore):
| File:<br>Waypoint | Description |
| - | - |
| US25:<br>US90_S<br>US90_N | If a non-exit-numbered highway is a cross road twice, add an underscored suffix. The direction letter refers to the relative position of the intersection along the route whose file is being made. US90_S is the southern of the two US 90 junctions along US 25, which runs S-N.

Duke87ofST:
Alright I swapped the directional banners and moved the southern point on MD373 over.

PR #3981, opened 2020-06-16
finst review + french/spanish openings

si404:


michihdeu:
@si404 all FRA-NAQ entries are E603 but should be N141 and N1141.

PR #3991, opened 2020-06-21
US175 N end relocation

yakra:
North end removed from S. M. Wright Freeway between TX 310 and I-45 Exit 283, and relocated onto C. F. Hawn Freeway between TX 310 and I-45 Exit 282.

Ping @ajfarina @ajfroggie @ovoss @jteresco
This will require changes to your .list files.


jteresco:
Well, that's annoying.  I no longer will have any travels on US 175!  Thanks for the heads up though.

PR #3996, opened 2020-06-24
WY 330 decommissioned

the-spui-ninja:
https://forum.travelmapping.net/index.php?topic=3692.0


rickmastfan67:
@the-spui-ninja You need to remove the 'wy.wy330' part in the updates.csv file.  Otherwise, it goes to a broken link, as we always leave that part 'blank' when removing a decommissioned route.

the-spui-ninja:
Ah okay thanks

rickmastfan67:
np. ;)  The only exception is if the route is completely renumbered (or merged).

PR #4017, opened 2020-07-01
Minnesota Updates - 7/1/20

ajfroggie:
Updates and corrections to several Minnesota lists, including route deletions.  Also includes relocating two points on the WI US 63 file (with Jeff Morrison's approval) that are the result of the Red Wing Eisenhower Bridge project.

### DIFF
--- a/updates.csv
+++ b/updates.csv
@@ -3485,6 +3485,14 @@
 2018-09-03;(USA) Minnesota;MN 101;mn.mn101ch2;Created separate MN 101 list in Chanhassen due to recent truncations resulting in a non-continuous route
 2018-09-02;(USA) Minnesota;US 169 Business (Shakopee);mn.us169bussha;Created route
 2019-07-30;(USA) Minnesota;MN 225;;Deleted route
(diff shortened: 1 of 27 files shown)


updates.csv
```diff
@@ -3485,6 +3485,14 @@
 2018-09-03;(USA) Minnesota;MN 101;mn.mn101ch2;Created separate MN 101 list in Chanhassen due to recent truncations resulting in a non-continuous route
 2018-09-02;(USA) Minnesota;US 169 Business (Shakopee);mn.us169bussha;Created route
 2019-07-30;(USA) Minnesota;MN 225;;Deleted route
+2020-06-29;(USA) Minnesota;I-90;mn.i090;Exit numbers revised from MN 76 to the WI Line
+2020-06-29;(USA) Minnesota;MN 3;mn.mn003;Route extended north to MN 5
+2020-06-29;(USA) Minnesota;MN 101;mn.mn101ch2;Deleted route
```
yakra, line 3490: `mn.mn101ch2`should be removed from the root field.

jteresco: I'll remove that after I merge.

PR #4078, opened 2020-08-02
Creating PA 423 TRUCK (Pocono Pines) and Minor PA 248 Edit

Markkos1992:
PA 248: Added Magnolia Dr (MagDr) and changed Valley View Rd (ValViewRd) to Valley View Dr (ValViewDr).

Also removed altername name for Future I-86 and alternate label for US 1.


Markkos1992:
Datacheck Successful after not thinking straight on here earlier.

PR #4113, opened 2020-08-15
NV I-11 exit renumberings

ovoss:
Also affects US 93 and US 95. All labels in use kept as alternate labels. No Updates entry needed.

Datacheck.sh successful.


rickmastfan67:
Heads up @ovoss, but the '56' label in US-93's file is listed as in-use.  Should be re-added as an alt label to 'I-11(17)' before this is pulled in.

ovoss:
> Heads up @ovoss, but the '56' label in US-93's file is listed as in-use. Should be re-added back as an alt label to 'I-11(17)' before this is pulled in.
> 
> https://github.com/TravelMapping/UserData/blob/91f86a146413344900551540a26ff8da39a76ee0/list_files/charliezeb.list#L1401

Fix made in new commit. Thanx.

PR #4121, opened 2020-08-18
BW

michihdeu:


rickmastfan67:
Update log entry is broken.  @michihdeu, you forgot the extra '0' in the file name for l200.  Because of that, it's leading to a broken link.  Same issue with l288 the day before.

michihdeu:
Thanks, I hate that I cannot test this beforehand......

PR #4124, opened 2020-08-20
CO, MT, WY fixes and cleanups

the-spui-ninja:
hopefully I got everything and nothing breaks


yakra:
MT US2 needs a point for the new SR327, with the old one becoming `CliSt`, and an updates entry noting that `SR327` has been moved.

The existing updates entry for SR327 itself isn't too clear on what alignment SR327 was removed from; it may not be immediately apparent what changes need to be made to a .list.

the-spui-ninja:
Alrighty; I’ll get to that once my internet problems get fixed (that should be in the next day or two hopefully).

the-spui-ninja:
OK am back after internet/semester/family business, will get to this today or tomorrow.

PR #4172, opened 2020-09-15
Colorado merges and deletions

the-spui-ninja:
hopefully I got all the csvs updated


si404:
I'll check the .csvs

si404:
they work, so that's a good sign

PR #4197, opened 2020-09-25
BB/NOR

michihdeu:


michihdeu:
B5 source is mauttabelle.de (not there since minimum Mid 2018 - official Brandenburg maps are also from Mid 2018 and still have B5 though)

PR #4203, opened 2020-09-28
KY: add 5 new routes, delete Future I-66; other minor KY, LA, & OH edits

mapcat:


rickmastfan67:
@mapcat The CumPkwy file would need to be updated first to include the 'in-use' labels from I-66Fut's file since it used named labels, while CumPkwy uses exit number labels.

`ky.i066futbow(30): 29(CUMPKWY) I-65 KY55 KY61 KY90 KY914 OLDCUMPKWY US27 US31E US68_EDM`
vs
`ky.cumpkwy(30): 1 11 14 15 27 29 46 49 62 78 86 88 OLDCUMPKWY`

@jteresco I would wait to pull that in till that's fixed, as it possibly could break list files.

mapcat:
Crap. I will do this now. It should only take about 20 min. Would be great to get this pulled in. Hope you can wait.

mapcat:
Thanks James

mapcat:
I think that's all of them. In the future, will these all show as in-use labels for Cumberland Pkwy?

rickmastfan67:
> 
> 
> I think that's all of them. In the future, will these all show as in-use labels for Cumberland Pkwy?

They should.  That's why I mentioned that the CumPkwy file needed them, since the I-66Fut labels would have been completely wiped from the database when the file is deleted, making the alt name for the CumPkwy completely worthless in the CSV file.

Now, if both files had exactly the same labels, just adding the alt name to the CSV file would have been all that was needed.

jteresco:
Looks like this is good to go, so I'm merging in for the update.  Thanks.

PR #4215, opened 2020-10-02
TUR O21 opening

rmihailucian:
O21 opening finally in.
@michihdeu please perform datacheck on this.
@jteresco please hold on this one until datacheck is confirmed to be ok


michihdeu:
@rmihailucian There was a conflict with updates.csv because both pull requests had edits in the same line. I think it is better to make one pull request in situations like this.

btw, what's about https://forum.travelmapping.net/index.php?topic=3732 ?

PR #4243, opened 2020-10-19
CA I-710 extension, other CA updates

ovoss:
Extends south end of I-710 to CA 47, over replacement for Gerald Desmond Bridge in Long Beach, per forum discussion. Also includes tweaks to other routes, some of them also discussed on forum.

Updates item in separate pull request #4244.

Datacheck successful.


rickmastfan67:
Shouldn't CA-1 be updated too since it's Concurrent with US-101 between 128 & 132? @ovoss

ovoss:
> Shouldn't CA-1 be updated too since it's Concurrent with US-101 between 128 & 132? @ovoss

Yes. Will do that separately, once this pull request is processed, to avoid interference with the pending request.

rickmastfan67:
rgr that.

PR #4247, opened 2020-10-20
OH & KY: route extensions, reroutings, & deletions

mapcat:


Markkos1992:
The "ky.ky3022" should be removed in the KY 3022 updates entry since the route was deleted.

PR #4259, opened 2020-10-27
ISL TH60 new alignment

Duke87ofST:


Duke87ofST:
@si404 there's also an update to usaush (az.us066hiswin) caught in here. Synchronizes waypoints with other routes in the area.

PR #4260, opened 2020-10-28
NY17 Exit 125 etc.

yakra:
https://forum.travelmapping.net/index.php?topic=3850


Markkos1992:
"ny.ny0272" was written for the NY 272 Updates Entry instead of "ny.ny272" leading to the link going to a blank page.

PR #4276, opened 2020-11-02
WA and ID updates

compdude:
Various updates to WA and ID:
-Fixing NMPs at border waypoints
-Removing some unused alt labels
-Renaming waypoint at west end of I-90 (old waypoint kept as alt label)
-Truncating ID 71 at northern end and adding updates entry for this change

### DIFF
--- a/hwy_data/ID/usaid/id.id071.wpt
+++ b/hwy_data/ID/usaid/id.id071.wpt
@@ -13,4 +13,4 @@ CmpCrkRd http://www.openstreetmap.org/?lat=44.747730&lon=-116.870780
 +X06 http://www.openstreetmap.org/?lat=44.763575&lon=-116.904761
 +X07 http://www.openstreetmap.org/?lat=44.804903&lon=-116.918701
 PowPlaAccRd http://www.openstreetmap.org/?lat=44.842206&lon=-116.894936
-ID/OR http://www.openstreetmap.org/?lat=44.845220&lon=-116.897112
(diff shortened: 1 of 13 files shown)


yakra:
Where does AtlSt come from? I see 4th Ave and Edgar Martinez Dr.
Everything else looks excellent.

Zoinks - I just had to check out the 2-state US195Spr in GMSV, and it appears to only be signed TO 95 / TO 195. Looks like this may not be a keeper... :(

compdude:
Oops, for some reason I forgot that it's called Edgar Martinez Drive there.

As for the US 195 spur route, I'm surprised that was even added to CHM/TM
in the first place. The Google Street view along the spur route dates from
~2008-9, and even that has a TO 195 sign on the Idaho side and no US 195
Spur signs. I guess this is not unlike Washington's other short spur routes
that aren't signed.

I will make these changes when I get home from work tonight.

yakra:
We went pretty heavy on the unsigned/apocryphal US Routes *(ME US1 AltBus (Bangor), anyone?)* in CHM's early days, then removed many of them. No big surprise that there are still a few kicking around.

jteresco:
Good to go?

compdude:
Not quite yet. Give me another hour or so.

jteresco:
No rush, I'll have it go in tomorrow instead.  Tonight's update has completed.

compdude:
Okay, it should be good to go now for tomorrow's update.

compdude:
Okay, _now_ it should be all good to go for the next update. Forgot to do the change to I-90.

PR #4307, opened 2020-11-13
usany truck route removal; NY340 relocation

yakra:
The original plan was to activate usanyp along with this, but [that's hit a snag](https://forum.travelmapping.net/index.php?topic=1928.msg20776#msg20776).

ping @ua747sp -- This moves the `HigAve` point on NY340, and will change how your .list file is interpreted.
If you have in fact traveled NY340 via Main & Union, then
`NY NY340 ValAve_S NY303`
If not; if you've done Valentine & Piermont, then
`NY NY340 ValAve_S ValAve_N`
`NY NY340 PieAve NY303`


ua747sp:
Thanks for the heads up. I've almost certainly traveled Main, but probably not Union. I'll keep an eye on things and update as needed when this update goes to production.

jteresco:
Removal costs me some mileage and some clinched routes, but it does make me happy to see them go.

PR #4320, opened 2020-11-17
Turner Diagonal diverging diamond

yakra:
With a new diverging diamond in place at I-70, now the Turner Diagonal is *really*...
not_a_freeway


the-spui-ninja:
Since you're looking at KCK anyway, do you think the freeway section of the Shawnee Mission Parkway (I-435 to Widmer Road) should be thrown in, based on things like the Sam Jones Expressway and Shadeland Avenue in Indy?

ovoss:
The freeway segment is only about two miles long.

There've been recent concerns on the forum about overdoing it on additions to the usasf system, and suggesting a moratorium on additions until there's more consensus about what belongs in usasf. Wherever we end up on that, I think the Shawnee Mission addition would be on the wrong side of the cut line.

Oscar

yakra:
Not something I'd be interested in adding. To my eyes, it's more of a boulevard that happens to have a few interchanges on it.
To add something to usasf, I'd want it to be an all-freeway facility, preferably something that forms a cohesive link in the larger network.

the-spui-ninja:
That's fair; then maybe those ones in Indy should be looked at with a view of keeping everything consistent.

yakra:
I'll leave that as @mapcat's call.
BTW, IN KeyPkwy is missing a point at Lowes Way, just south of the N terminus.

PR #4342, opened 2020-11-25
UKR/FIN/ENG/MKD

si404:


si404:
@michihdeu - I broke your (and only your, as alt labels/route names covered everyone else) .list file in the deletion of the non-motorway eursf routes in Finland, because you've travelled on the St103 bit of Keha III and have no entry for that other than the Keha III one.

michihdeu:
Thanks, I'll check it tomorrow.

michihdeu:
@si404 Moottoritie Turku-Lieto HB link does not work

PR #4350, opened 2020-11-28
Patch 7

ajfroggie:
Reposition the I-85 AL/GA point to the mean water mark on the Alabama side, which per court records and a forum discussion is the official AL/GA state line.


yakra:
ping @ntallyn

PR #4365, opened 2020-12-04
New South Wales

si404:


si404:
This takes us to 400 systems active+preview (286 active, 104 preview).

michihdeu:
4**1**0 (2**9**6 active, 104 preview)

si404:
400, but yes I did mistype with 8 rather than 9.

michihdeu:
oh.... I just remembered that we have 296 and.... I think I'm better in typing than calculating.... 😆

PR #4433, opened 2020-12-28
fix for #4124

the-spui-ninja:
@yakra does this work? Or do I need to be more descriptive in the update section?


yakra:
https://travelmapping.net/devel/manual/maintenance.php#newsrelend
> * **Relocated route (at end of route):**
`Removed from Main Street and 5th Avenue between 4th Street and 9th Street, and relocated onto a new northern Georgetown bypass between 4th Street and PA 70.`
Mentions both the new and old routings as well as the bounding intersections of each.

https://travelmapping.net/devel/manual/maintenance.php#newsavoid
> * **Avoid the following in the entries:**
>   * Don't refer to the `old route` or the `old end`. Instead, say what the old route or end was.
>   * Don't refer to the `new route` or the `new end`. Instead, say what the new route is.
>   * Don't use vague phrases like the `correct location`, `onto new construction`. Instead, say what the location or highway is.
>   * Wherever possible, avoid describing old and new routings by highway names that changed. Use stable highway names if they are available, or describe the type and location of the highway.

When users need to change their .list entries as a result of a Highway Data update, they can only see the new version of what's in the HB, but their .list reflects the old version. Thus we must rely on updates entries to describe what's changed and how.

More descriptive updates entries help travelers in several ways:
* If a user needs to de-clinch something, they'll have a better idea of what the new alignment is and where it ties into the old.
* Knowing what the old alignment is, they can check if it's in the HB as something also now, and .list that if needed.
* Oftentimes a relocation will cause an in-use label to move to a different point. We need to know what to replace the old `MT US2 SR237` with.
  * To that end, since `MT US2 SR237` was in use and moved, a US2 updates entry would be appropriate here.

Adapting from the template above, I'd suggest something like:
```
2020-08-20;(USA) Montana;SR 327;mt.sr327;North end removed from Bainville-Snowden Road, 5th Avenue and Clinton Street and relocated onto a north-south road from Bainville-Snowden Road to US 2.
2020-12-27;(USA) Montana;US 2;mt.us002;Waypoint SR237 moved from Clinton Street to a new north-south road located 2.7 mi east.
```
(Here, we don't have a stable name for S-237 other than S-237, so the "describe the type and location of the highway" bit kicks in.)

the-spui-ninja:
#4482

PR #4478, opened 2021-01-12
GES

michihdeu:


si404:
Pretty sure Roman numbers _are_ Latin - you mean Arabic (really Indian) ;P

But yes, they are easier to read. I'm surprised Spain went with them for N1-6.

michihdeu:
yeah, I was not sure about the name... both is used on signs, GM, OSM,.... it sucks. I thought I should go with Roman numbers because they are more common but when I made the first wpt file frages.d0021biii.wpt - no!

PR #4519, opened 2021-01-30
canon: ON-66 bypass of Virginiatown

rickmastfan67:
See: https://forum.travelmapping.net/index.php?topic=4074.0


rickmastfan67:
@ovoss This edit might affect you depending on when you where last on ON-66.  See forum post.

PR #4540, opened 2021-02-07
California updates

ovoss:
Datacheck later today, especially with all the .csv edits.

As discussed in forum:

- Remove US 101 Business (Ukiah), and US 50 Alt (Pollock Pines)

- Remove CA 146 (Pinnacles NP), and CA 14 Truck (Newhall), with latter becoming an alternate route name for the concurrent I-5 Truck (Newhall)

- Add to usanp Mormon Emigrant Trail, and keep with tweaks Minaret Summit Rd

- Truncate I-305 to eastern US 50/CA 99 interchange

- Peer review edits to CA 59, CA 73, CA 98, CA 99, CA 140, and CA 186, and spot changes to CA 17 and CA 70

PR #4541, opened 2021-02-07
Update updates.csv

ovoss:
For changes made in PR #4540.


michihdeu:
@ovoss `Truncated from waypoint 7C to waypoint 6` I think `exit` should be used instead of `waypoint`, shouldn't it?

ovoss:
Agreed. Fix made.

PR #4547, opened 2021-02-09
split CT ferry routes

yakra:
For consistency with the sitewide practice of excluding ferries.


michihdeu:
@yakra are there still more cases in other regions?

yakra:
Nothing I know of in any regions I maintain.

PR #4594, opened 2021-03-04
DE: US 13 and DE 1 

Markkos1992:
US 13: Added Roosevelt Ave (RooAve), Rose Bowl Rd, Anchor Inn Rd (AncInnRd), +X3AB, +X3AA, and Governor Printz Ext (GovPriExt).
DE 1: Added Bakerfield Rd (BakRd), +X1B, +X00E, +X00D, +X00C, +X(DE6), +X00B, +X02A, +X3AA, and +X3B.


Markkos1992:
Probably.  I do not mind admitting to that I probably get too particular with that.  This started with eme looking into the DE 1 shaping point north of Five Points that I only have because the line otherwise follows Nassau Rd.

michihdeu:
> Isn't some of this stuff overdoing it, like the US13 `+X00\*` (E in particular) points? The centerline is already within tolerance, and the trace is sufficently delineated from nearby routes for mapview

Those shaping points are unnecessary. It is definitely too much! No benefit, just clutter which leads to higher loading times.

Markkos1992:
I thought that the affect on loading times at this point was minimal. (at least compared to the CHM days)

michihdeu:
> I thought that the affect on loading times at this point was minimal. (at least compared to the CHM days)

Sure, adding just a few points for less routes to avoid a sharp angle error or to improve distinguishing different non-concurrent routes is fine. If you plan adding points to more routes like you did here, especially mapview loading times will increase....

I think there is more important stuff to do for TM, e.g. more than 100 preview systems to be reviewed etc.

jteresco:
Extra points are definitely not as big a concern as they were in the CHM days.  But I think we do want to be fairly consistent and not shape when not needed.

I'll merge this in and get it into the update - can always thin out the points later.

Markkos1992:
Well in general, I tend to be cautious on how I do things.  I would personally rather not be near the tolerance lines.

Having stated that, I think my turn towards adding potentially too many shaping points started with I-579’s short length, and roads like PA 906 (which without shaping points had its line cross the Monogahela River).  I think I went maybe from one extreme (get rid of them) to another (be too particular) here.

Mark Moore

Civil and Environmental Engineering
Virginia Tech Class of 2014
markm10@vt.edu
(804)691-4381

yakra:
@jteresco wrote:
> Extra points are definitely not as big a concern as they were in the CHM days.

On the DataProcessing side, at least? We have the quadtree to speed up colocation detection; ISTR some posts focusing on that, in contrast to CHM potentially having had some O(n^2) operations.

@michihdeu wrote:
> especially mapview loading times will increase

This is what I'm thinking of. Still not O(n^2), but with coords all being transferred and processed as text, this can add up.

Markkos1992:
As a test, I just did my mapview related to my recent route planning for my trip yesterday (though I did not enter WV).  This is for WV, MD, DC, VA, and DE.

I did it twice and got about 42 seconds each time. (https://travelmapping.net/user/mapview.php?units=miles&rg=WV,MD,DC,VA,DE&u=markkos1992)

PR #4615, opened 2021-03-15
KY: US 68 relocation east of Millersburg

mapcat:


Markkos1992:
I think that there may be a typo in your US 68 updates entry.

mapcat:
Thanks for catching those. Fixed in next pull request.

PR #4618, opened 2021-03-16
ENG + SCT

si404:


Markkos1992:
Should the updates say "Route reentered" instead of "Route rentered"?

PR #4675, opened 2021-04-11
WV cleanup

rickmastfan67:
Cleanup US-52(Ken) so it plays nicer with KY-3 & KY-292 on the other side of the river border in places.  More points than I normally would do, but since the border was there & a river, tried my best to keep it over on the WV side now with the additional points.

Also added a few shaping points (or normal points) here and there to several other US routes to flush them out a little better where they were close to being out of tolerance.

No update mentions needed, as there were no reroutes.


Markkos1992:
@rickmastfan67 It looks like this created 3 sharp angle errors if you did not catch them already.

rickmastfan67:
@Markkos1992 I knew this was going to happen.  Just was waiting for the page to update before I fixed them.

Markkos1992:
I figured.  I have noticed that there are still a lot of leftover datacheck errors still around.

PR #4698, opened 2021-04-21
Patch 5

ajfroggie:
Corrections to the US 31/US 280 concurrency in Birmingham, AL and exit number changes to I-20/I-59.


rickmastfan67:
Don't forget to deal with I-90 in MN. ;)

PR #4724, opened 2021-05-02
JPN + ENG

si404:


Markkos1992:
The dates on the Updates Entries seem to be "2-5-2021" instead of "5-2-2021".

si404:
you are right, flipped it round for no good reason, the American is rubbing off!

PR #4757, opened 2021-05-11
WY 370 realignment

the-spui-ninja:
couldn't get anything else in before my flight leaves


the-spui-ninja:
That's smart, didn't think of that

Duke87ofST:
I'm updating my list to show my clinch of the new alignment anyway (since I drove both old and new last summer when I was scouting the signage situation), but thanks for the alert that it's time to do that. :)

TXStormchaser:
Updating mine as well. Thanks!

PR #4758, opened 2021-05-11
WY 370 alt points added and hopefully the I-76 BL in Colorado

the-spui-ninja:
Don't merge this for a couple hours


the-spui-ninja:
OK I should have the I-76 BL/BS stuff in. Did I do the alt route label in the CSVs right? I just went off the other ones I found.

the-spui-ninja:
Last I saw was that it was going to be a loop (since most of the signage is BL not BS) based off of Pine Bluffs

PR #4784, opened 2021-05-19
POL + JPN + WLS + sync

si404:


michihdeu:
@si404 POL: entended -> extended

PR #4809, opened 2021-05-25
usafl: FL-46 (split in two, + reroute)

rickmastfan67:
So, it seems that FL-46 has been rerouted onto a segment of Wekiva Parkway between 'FL429' (using Exit 44) & 'SleBearLn' (Old CR-46A).

It also seems that it's been split in two as well, as the segment between the current labels of 'CR46A' & 'FL429' has been resigned as CR-46A in the field & there's an 'END' shield @ the intersection where the current 'FL429' point is.

Sadly, anyway you look at this, several list files will need to be broken due to the route needing to be split into two due to the CR-46A segment.  Now, the GIS (dated 05/22/21) still shows the route as 1 piece (including CR-46A weirdly as SR-46A too), but I think they were planning on dropping the part between FL-453 & FL-429 anyways once all this construction was done.  So, since we have that END shield, CR-46A is posted, I can now at least split that CR-46A part out and deal with other parts when that time comes & signage is a little more complete.

====

People that will have their list files broken that update via email:
charliezeb, skatcher, flroads, htm_duke, klundgren, johninkingwood, 1995hoo, aaroads, synkdorbit, mefailenglish, presidentman (1 entry will be a new route, other needs double checked due to reroute)

Update via GitHub:
@Duke87ofST, @ovoss (reroute only part, list not broken, but will need updated), @jteresco (entry going to be on a new route), @freakwentflier, @ntallyn (1 of 2 entries will be broken), @thefxexpert, @neroute2, @pianocello130

People unaffected:
@ssoworld, M3200, dharwood, @mapcat, morriswa, rlee, @mapmikey, dfilpus, brendan, hsford

Please discuss any issues with this here: https://forum.travelmapping.net/index.php?topic=4338.0


rickmastfan67:
Ooops, messed up on the URL for the news entry for the 'new' file.  Fixed in #4815.

PR #4817, opened 2021-05-26
TX US180: add RicWilHwy

yakra:


Markkos1992:
Looking at Mapview, US 180, just before its eastern end, is shown as going south of I-20.  It is probably best to add Hudson Oaks Dr (HudOakDr) to US 180 to take care of this discrepancy.

yakra:
Oops, didn't catch on that RicWilHwy is a duplicate label. Will fix and also add HudOaksDr.

Markkos1992:
Well it is always nice to solve multiple issues at once.

PR #4824, opened 2021-05-27
IN: I-65/I-90 graph connection + new exit on I-65

mapcat:


jteresco:
I'm going to hold off on this as it looks like there's some additional discussion on the forum that might led to a different solution.  @mapcat @Markkos1992 @rickmastfan67

rickmastfan67:
Adding the link to the thread for reference:
https://forum.travelmapping.net/index.php?topic=4347.0

mapcat:
If you don't like the solution to I-90/I-65, fine, I will look at that when I have more time to do the research. In the mean time, can you pull in the other commits separately? Thanks.

Markkos1992:
@mapcat Am I correct that there should be an updates entry for KY 3538 in the meantime?

@jteresco Is there a way to just get rid of the last second in.i065 commit and the datacheck.fps commit easily?

mapcat:
@Markkos1992 probably not because (a) I am the only user who has been on it and (b) the file was added only 2 months ago with only a guess at the east end. The change is just a correction; the road was not extended.

jteresco:
I should be able to merge in just a subset of the commits.  It's something I want to figure out how to do anyway so now's as good a time as any.

So we want everything except https://github.com/TravelMapping/HighwayData/pull/4824/commits/0420889d942e1ba8b3fa1050b779204a92f92d68 and https://github.com/TravelMapping/HighwayData/pull/4824/commits/927d7efe065aaf700e20558194ac514c430ba8f0 if I'm understanding correctly.

mapcat:
@jteresco thanks, yes, those seem to be the only contentious items.

yakra:
ISTR @jteresco has `cherrypick`ed commits before.

jteresco:
I believe I have successfully cherry-pick'ed the commits we want from this.  @mapcat @yakra @Markkos1992

PR #4873, opened 2021-06-08
OCC

michihdeu:


michihdeu:
- Source for **N100** truncation: GSV (last N100 sign at roundabout westbound), GSV (mile post with D6100 halfway between roundabout and former end junction, wikisara says that the end is anywhere close to this mile post.
- Source for **N126** truncation: https://routes.fandom.com/wiki/Route_nationale_française_126
- Source for new **N524** route: https://routes.fandom.com/wiki/Route_nationale_française_542

PR #4877, opened 2021-06-09
Update nmpfps.log

ovoss:
For FP NMPs created by proximity of I-80BL (Reno) and I-80 junctions with NV 659.


yakra:
I-80BLRen@I-80_E looks like it should be synced to the other points
http://www.openstreetmap.org/?lat=39.533306&lon=-119.738660

ovoss:
Thanks! That fix should eliminate the NMPs relating to I-80(19)/I-80BL/NV659. Closing this PR accordingly.

PR #4921, opened 2021-06-23
whole boatload of Colorado updates (per forum and per personal field-checks)

the-spui-ninja:


the-spui-ninja:
let me know if there are any issues

Markkos1992:
Is CR90_W a correct label for the west end of CO90Mon?

the-spui-ninja:
Yeah, the sign at the end of the route says “End State Highway 90, begin County Road 90”

yakra:
So CR90 begins there, rather than continuing?

the-spui-ninja:
oh yeah fair point; should have that taken care of now

PR #4999, opened 2021-07-29
CO/MT updates

the-spui-ninja:


the-spui-ninja:
The co.co040lim.wpt deletion is a holdover from last year; I removed the route from the .csvs and made the update entry but forgot to remove the actual file

PR #5086, opened 2021-09-01
OH: point label fixes

mapcat:


michihdeu:
@mapcat Point label changes are not newsworthy. You can use alt labels for the wp if the point is in use.

Markkos1992:
@michihdeu I reported the change originally.  It was necessary in this case due to how OH 113 was previously in the HB.  (The west end had the OH269 label that needed to be moved.)

PR #5121, opened 2021-09-21
Create jpn.n131.wpt

denishanson:
Minato (port) national highway


denishanson:
Will resubmit after testing

PR #5131, opened 2021-09-23
Initial jpnh submission

denishanson:
System files and minato (port) national highways wpt files for the **jpnh** Japan National Highway system. Minato highways are very short national highways that connect seaports/airports to other national highways.


denishanson:
Verified using the data verification shell script

PR #5138, opened 2021-09-26
Systems jpnhhk and jpnhks

denishanson:
Added N338 and devel-only systems jpnhhk (Hokkaido) and jpnhkt (Kanto)


denishanson:
Ran datacheck

michihdeu:
@denishanson Why have you changed the system color? Blue is already used for the tier 1 system. I made a [note on the forum thread](https://forum.travelmapping.net/index.php?topic=4248.msg25424#msg25424). Please check it and considered it for drafting new routes. Thanks.

denishanson:
I am color blind and have trouble seeing the lighter colors.  I'm using
blue while I develop the national highways so that I can easily see the
routes I added. I plan on changing the color on jpnh to something
appropriate when it is promoted to preview.  The other systems are dev-only
and will be deleted once everything is pushed to jpnh.

denishanson:
Would red be a better, temporary, color to use? I can see it better than I
can see teal, yellow, or salmon.

michihdeu:
@denishanson the color doesn't matter as long as the jpnh systems are in devel. You should change it when the status will be changed to preview.

PR #5148, opened 2021-09-28
All Okinawa Prefecture national highways

denishanson:
All Okinawa Prefecture national highways.   The smallest prefecture.


denishanson:
Successful datacheck

PR #5149, opened 2021-09-28
Update hun.f001.wpt

rmihailucian:
Changes to F1 requested in 5146.


michihdeu:
@rmihailucian You have two F13 labels now (duplicate label error). The new one must get a different name. I suggest F13_Kop or F13_W

PR #5162, opened 2021-09-30
Glenwood Springs bridge complete + CO alignment updates

the-spui-ninja:


ovoss:
I can't figure out, from the Updates entry or the pull requests above, how CO 82 was changed in Glenwood Springs, and in particular the waypoints between which the change happened (so I can edit my list file).

PR #5177, opened 2021-10-03
Operation NYCR part 3

yakra:


Markkos1992:
Should we add a point [here](https://www.google.com/maps/@43.1140585,-75.2026468,3a,75y,342.14h,92.26t/data=!3m6!1e1!3m4!1ssv4pum9yZlxJVMACfh6c7A!2e0!7i16384!8i8192) on NY 5 while we are at it?  It looks like traffic to I-90 is directly to go a different way than using I-790/NY 5.

yakra:
`HerHd_W` added.

PR #5197, opened 2021-10-10
usams first batch

ajfroggie:
First batch of usams routes for development.


yakra:
Do you want to add a line to [systems.csv](https://github.com/TravelMapping/HighwayData/blob/master/systems.csv) to add usams to the HB as a devel system, are are you just uploading the wpt files for now?

ajfroggie:
I made a post that they were pulled, so should probably include the line.

PR #5221, opened 2021-10-16
Long Island fixes / NYCR part 8

yakra:
Operation NYCR is now just over 50% complete. Carl Rogers eat your heart out.


yakra:
* [x] NY212 -> NY212_W
* [ ] A graph connection should be added with I-87. This is what Jason, who originally drafted usact, usari & usanj, would have called "a bit rampy". The example from my states I always think of first is [NH US3 & pals @ I-93(20)](https://travelmapping.net/hb/showroute.php?r=nh.us003&lat=43.453168&lon=-71.570821&zoom=15). The layout is a bit more like [US1 @ US7](https://travelmapping.net/hb/showroute.php?r=ct.us001&lat=41.118111&lon=-73.419893&zoom=15), though the execution is a bit different, with the NY212 split begging for a point where NY32 turns the corner. (Right here on NY32, compare [I-84](https://travelmapping.net/hb/showroute.php?r=ny.ny032&lat=41.519471&lon=-74.020343&zoom=16).) `I-87_N` will become an AltLabel of the new `I-87` point (also on NY212), and `I-87_S` will be collapsed into `NY212_W`.
* [ ] `US9W_Bet` and`US9W_Ken` should become `US9W_BetS` and `US9W_BetN`. Similar for US9W itself.
* [ ] Check over the rest of the 9W/32 junctions too.

PR #5267, opened 2021-10-29
PA Revisions 10-29-2021

Markkos1992:
I am one one of those look at things moods so there will probably be more added later.

I-70:  Shaping Point Revisions
I-76:  Added future PA 130 Interchange (https://www.post-gazette.com/news/transportation/2021/10/28/Turnpike-to-build-interchange-in-Penn-Township-Westmoreland-County/stories/202110280153) (https://www.aaroads.com/forum/index.php?topic=419.msg2676330#msg2676330). Also shaping point revisions.
I-376/US 22:  Shaping Point Revisions
US 30: Added Pleasant Hills Rd (PleHilRd), Century Farm Rd (CenFarmRd), and Raccoon Park Rd (RacParkRd).  Also shaping point revisions.
PA 50: Added Bell Ave.
PA 88:  Replaced +X7 with Kirchner Dr (KirDr). (https://forum.travelmapping.net/index.php?topic=4652.0)
PA 168:  Added Temple Rd (TemRd) and Townsend Rd (TowRd).


yakra:
Take care to not add too many shaping points. If a route is already within tolerance in wptedit, there's probably no need to add additional points unless they see a significant amount of turning traffic.
PA has 17674.50 active/preview miles and takes ~49 s to render in mapview.
In contrast, neighboring OH has 19407.30 active/preview miles and takes ~26 s to render in mapview.

Markkos1992:
Definitely the one downside of myself being picky with the red lines.

yakra:
[be conservative. Not every curve needs a shaping point. Few curves ever need more than one shaping point.](https://travelmapping.net/devel/manual/includepts.php#sharpturns)

Markkos1992:
Please pull this in as-is @jteresco.  I do want to relook at things as I move forward, but I am good with it for now.

rickmastfan67:
Then again @yakra, PA is SUPER curvy compared to Ohio. ;)

Also @Markkos1992, just me, but I would have avoided putting a hidden point for PA-130's future interchange on I-76 at this time.  Why you ask?  Because as far as I know, they haven't decided on the configuration of the interchange yet.  It could be setup similar to a 'normal' Turnpike interchange still due to all the roads in close proximity of where PA-130 crosses over.  Especially since they have to factor in the 3-laning there too.  But as I said, that's just me.

Markkos1992:
@rickmastfan67 Yes, I am aware that the point will most likely have to be relocated as the configuration has not been decided yet.  I would rather it be there than not though, but we can agree to disagree on this one.

PR #5269, opened 2021-10-30
US 30: Point Revisions

Markkos1992:
Replaced +X104 with Helixville Rd (HelRd).  Replaced +X104A with Sleepy Hollow Rd (SleHolRd).  Added Graceville Rd (GraRd).  Adjusted and removed shaping points.


Markkos1992:
I realized that the US 30/US 30 BUS TRUCK/US 322 TRUCK shaping point was there to prevent sharp angle errors.

PR #5273, opened 2021-10-31
I-69C/E exits

yakra:


Markkos1992:
More of a question than anything else, but are town names used on US 77 north of I-69E mainly because there are no better labels to place here?

yakra:
Short answer: Yes.

These points were added to CHM to break up visible distance segments, and there were no road names available at the time.
These days, I *do* see road names in ESRI, but these aren't signed. I can't corroborate them in any official sources -- these roads aren't shown in shapefiles, the statewide planning map, or county grid maps. Nothing.
So I'm just leaving them be. This'll all be converted to I-69E eventually anyway. ;)

PR #5330, opened 2021-11-15
WV: US-220 corridor cleanup

rickmastfan67:
https://forum.travelmapping.net/index.php?topic=4572.0


rickmastfan67:
Also added in the new WV-14/31 roundabout mentioned here: https://forum.travelmapping.net/index.php?topic=4562.0

PR #5384, opened 2021-12-01
extend L55K

yakra:
https://forum.travelmapping.net/index.php?topic=4699


Markkos1992:
I just realized that the file root is not being shown in the L55K updates entry.

yakra:
https://github.com/TravelMapping/HighwayData/pull/5403

PR #5408, opened 2021-12-08
KY & OH: extend KY 1679 & add US 30 Bus (Van Wert, OH)

mapcat:


Markkos1992:
KY 1679 was written was KY 1697 in the updates entry.  It is pretty easy to catch when I am linked to the wrong road.   (sadly I do not see a laughing emoji here)

PR #5439, opened 2021-12-20
POL+CL

si404:


si404:
@michihdeu Portugal border point for extended A25/diverted E80 at
PRT/ESP http://www.openstreetmap.org/?lat=40.613293&lon=-6.819570

PR #5445, opened 2021-12-22
POL+ENG

si404:


si404:
The Poles made a right mess of that opening (some today, some tomorrow, some not yet, different bit partially opening early). Have time on the PC tomorrow to deal with this, the surprise early A1 and S19 openings. Plus Finland.

Closing until it can be done right.

PR #5487, opened 2022-01-04
TXSpr376 & TXSpr377

yakra:


yakra:
I just spotted two new routes to add to usatxs.
I should be able to get these in pretty quick, but if it's almost time for tonight's update, no big worries.

yakra:
Ready.

PR #5510, opened 2022-01-09
Hokkaido all remaining routes

denishanson:


denishanson:
Passed datacheck

denishanson:
Passed datacheck

PR #5565, opened 2022-01-31
Promote Tohoku roads in jpnh to preview

denishanson:
53% of Japan national highways are now in preview


denishanson:
Passed datacheck

PR #5588, opened 2022-02-09
add TXSpr29, delete TXLp118, etc.

yakra:


yakra:
Extraneous `tx.lp0118` will be removed next update.

PR #5622, opened 2022-02-19
BY/SN

michihdeu:


michihdeu:
@rmihailucian you want to update your SN S190 entry! The updates.csv description was difficult because it is all "Alte Altenberger Straße". You simply have to replace "AlteAltStr_S" by "B170_S" (maybe) or "AlteAltStr_N" (most likely) if you fully drove through Dippoldiswalde.

PR #5624, opened 2022-02-20
usapr preview -> active

mapcat:


mapcat:
Data check successful

Markkos1992:
I noticed that updates entries are missing for IN 750 and LA 1112.  I think I was able to quickly realize that these changes not affect my current (or future travels) though.  The LA 1209 removal actually makes my future clinching life easier.

mapcat:
Those updates weren't necessary because IN 750 disappeared on 2019-01-01 and I failed to delete the file when I removed the it from the .csvs. LA 1112 was identified in peer review as possibly unsigned, and never made it into the .csvs even though the file was there in case I learned otherwise. I noticed it in the list of unprocessed .wpt files when I scanned it for potential usapr issues.

Markkos1992:
Fair enough.  I will defer to others on if anything else is needed.

Mark Moore

Civil and Environmental Engineering
Virginia Tech Class of 2014
***@***.***
(804)691-4381

PR #5630, opened 2022-02-21
Removing PA 420 ALT TRUCK FP

Markkos1992:


yakra:
https://forum.travelmapping.net/index.php?topic=4716.msg27030#msg27030
> That shows up as Crossed Off (rather than [unmatched FP](https://travelmapping.net/logs/unmatchedfps.log)) because it's still flagged as an error, on NO_ABBREV grounds, but the .wpt data hasn't been changed.
This is a rather tricky case; it gets that flag because the underscore is present, so what we have is an underscored suffix rather than an abbrev.
>
> Removing the underscore and relabeling to PA420AltTrkPro, thereby matching the .list name of the intersecting route, will make the error go away.

Markkos1992:
I fixed the label.

PR #5636, opened 2022-02-24
NY I-278: recenter NJ/NY

yakra:
and delete NY340 PieAve; was mentioned on the forum somewhere


Markkos1992:
NY 340 reference:  https://forum.travelmapping.net/index.php?topic=4821.msg26944#msg26944

PR #5649, opened 2022-02-28
HUN M4 opening Szolnok Bypass

rmihailucian:


Markkos1992:
The updates entry should be updated to the following:

2022-02-28;Hungary;M4 *(Cegléd)*;**hun.m004ceg**;Route extended at eastern end between exits 93 and 117

PR #5734, opened 2022-04-22
ROU DX12 opening

rmihailucian:
@michihdeu please add a directory and move the .wpt file there. Also new systems files have to be created. Added also the section that is supposed to be finished in 3-4 weeks from now, I hope it's ok.


michihdeu:
@rmihailucian I'll remove the update entry and replace it by a system update entry. How should the new system be called? `Romania Drumuri ???`

rmihailucian:
@michihdeu Thanks a lot, the name should be Romania Drumuri Expres

michihdeu:
@rmihailucian You called the wps `DEX12` but the route `DX12`. I'll call both `DEx12` with E and lowercase x as indicated on wikipedia and on signs according to videos, e.g. https://www.facebook.com/video.php?v=501191545044608

Fine to you?

rmihailucian:
@michihdeu yes, that's fine

PR #5742, opened 2022-04-28
All remaining Kanto 300 and 400 roads

denishanson:


denishanson:
Passed datacheck

denishanson:
Passed datacheck

PR #5743, opened 2022-04-29
Promote jpnhkt to preview

denishanson:


denishanson:
Passed datacheck

michihdeu:
@denishanson please undo the systemupdates.csv changes! The old entries were fine. We have only ONE region for Japan. Japan is not split into regions on TM!

denishanson:
Reverted "(Japan) xxxx" back to "Japan".  Kept the promotion of jpnhkt to preview.

PR #5745, opened 2022-04-30
GES/NAQ/NOR

michihdeu:


michihdeu:
@rmihailucian You are the only traveler of the Norwegian route (Fv17 / Helgelandskysten Nasjonal Turistvei) changed with this pull request. Your Romantische Straße [log file error](https://travelmapping.net/logs/users/panda80.log) is caused by [this change](https://forum.travelmapping.net/index.php?topic=367.msg27141#msg27141). You also still have traveled the [B10 Rosenstein tunnel in Stuttgart](https://travelmapping.net/hb/showroute.php?units=miles&u=panda80&r=deubw.b010) which was just opened late March. Is that correct?

PR #5746, opened 2022-04-30
PDL

michihdeu:


michihdeu:
hmm, the other way round. Sync IeV + May border points with MeL shp files 😆

PR #5782, opened 2022-05-16
AZ: US 191 truncation, US191Bus creation, US191Spr deletion

Duke87ofST:


Markkos1992:
I had actually noticed last September that US 191 Spur seemed unsigned, but I never thought about relooking further into it.

PR #5796, opened 2022-05-18
Activate fracord and norfv

si404:


michihdeu:
@si404 The activation dates should be 2022 😄

michihdeu:
I'm going to merge and change it now

si404:
I think I went "oh, that's not quite a year old, I don't need to change the month" for Corsica and then didn't change the year either. And then copied the date for Corsica for Norway. 🤦 

Any way. Systems completed 🥂

PR #5822, opened 2022-06-02
All remaining Kansai roads

denishanson:
N372-N376
Renamed and moved wps
Removed Old and Spr from wp names


denishanson:
Passed datacheck

PR #5831, opened 2022-06-05
Ontario NMP FPs

rickmastfan67:
Thought I'd at least push this very quickly for you Jim, even though SH-538 is still part of the "in development" Secondary highways set for Ontario.


jteresco:
Thanks.

PR #5833, opened 2022-06-05
Merge all Japan national highways into jpnh

denishanson:
Merged jpnhhk, jpnhks, jpnhkt, jpnhky, jpnhsh into jpne


denishanson:
Passed datacheck

PR #5860, opened 2022-06-15
More Japan NMP fixes and data error FPs

denishanson:


denishanson:
Passed datacheck

denishanson:
Passed datacheck

PR #5878, opened 2022-06-24
Merge N339Kai and N339Sot into N339

denishanson:
Two drivable sections separated by a set of stairs (N339Kai)


denishanson:
Passed datacheck

PR #5913, opened 2022-07-12
ND: add pointless Bus 13 that's signed for some reason

the-spui-ninja:
double checked should be good this time
Will leave the involved MT change for later in the week when I have more time


the-spui-ninja:
Added MT because I had more time today

jteresco:
Site update failed:

<pre>
[454.1] ABORTING due to 1 errors:
1: Duplicate main list name in mt.sr201: 'MT SR201' already points to mt.mt201
</pre>

`MT SR201` is the .list name for `mt.sr201` but is also an `altRouteName` for `mt.mt201`.

Tonight's update is already several hours late, so I'll just leave this until morning, hope for a fix from @the-spui-ninja fairly soon and will re-run then.

the-spui-ninja:
I was worried something like that would happen; thankfully it’s only vespertine that’s affected. Guess there isn’t a way to keep travels if a route is split (not like this is a common problem though so don’t worry about coming up with a way).

jteresco:
There are tricks people use with alt labels to preserve some travels but with only one user affected, and that being one who updates regularly, it might not be worth the trouble.  Thanks for the quick fix.  I'll try the update again now.

the-spui-ninja:
Well and how many cases are there of a route getting partially "upgraded"? This is probably the only one we'll see, unless MDT does it again...

PR #5916, opened 2022-07-13
fix for #5913

the-spui-ninja:
Sorry about that, thought I could link the upgraded route with the old route


rickmastfan67:
Sadly, alt names only work if there isn't a route with the same name still in the database.

PR #5919, opened 2022-07-14
reviews and openings

si404:


michihdeu:
@si404 `2022-07-15;Montenegro;A1;mne.a001;Route added` is not necessary since it is a new system you already notified `2022-07-15;Montenegro;mnea;Montenegro Autoput;active`

additional more notes:
- I'd like to have a look at all review systems before activation
- The popped up UK/IRL systems need explanation: signed?, dedicated, tier classification,...

si404:
> @si404 `2022-07-15;Montenegro;A1;mne.a001;Route added` is not necessary since it is a new system you already notified `2022-07-15;Montenegro;mnea;Montenegro Autoput;active`

Yes, but that new system will be buried under a load of other activated systems once you've reviewed them (and I've not yet implemented all your notes, so there's a system or two more coming).

> * I'd like to have a look at all review systems before activation

Absolutely fine.

> * The popped up UK/IRL systems need explanation: signed?, dedicated, tier classification,...

Yes, yes, and 5 is probably right as we haven't gone 6 deep.

michihdeu:
> 5 is probably right as we haven't gone 6 deep.

I don't agree. If we need to go deeper, we should go deeper.

michihdeu:
> Yes, but that new system will be buried under a load of other activated systems once you've reviewed them 

So what? We face the same "problem" with every update.....

si404:
> > 5 is probably right as we haven't gone 6 deep.
> 
> I don't agree. If we need to go deeper, we should go deeper.

We don't need to - the systems I've added are comparable with other tier 5 systems. If these needed tier 6, we'd have done it already.

> > Yes, but that new system will be buried under a load of other activated systems once you've reviewed them
> 
> So what? We face the same "problem" with every update.....

True, but the 'A1 new route' is the more relevant info than 'mnea new system'. I see no reason to get rid of either.

michihdeu:
`irl.r1dub` is an invalid link. Should be `irl.r802dub`

PR #5951, opened 2022-08-04
WY Yellowstone updates and WY 346 creation

the-spui-ninja:
So I was able to get around this manually, but I got this error when updating the CSV's: 
image
Is there a way around this for the future?


michihdeu:
> Is there a way around this for the future?

Just open any csv file and edit the url in the browser address bar.

the-spui-ninja:
OK that’s what I did. I just was wondering if there were any GitHub tricks I didn’t know about, since the search I did on this had a lot of jargon I didn’t understand.

PR #5959, opened 2022-08-11
irlorb review

si404:


michihdeu:
@si404

`2022-08-11;Ireland;R801;irl.r801;Route extended at west end from J40 to R105/R138` should be read:
`2022-08-11;Ireland;R801;irl.r801;Route extended at west end from J41 to R105/R138`

Maybe: 
`2022-08-11;Ireland;R105 (Dublin);;Route deleted` change to
`2022-08-11;Ireland;R105 (Dublin);;Route merged into main route` ? I mean, I just got a `deprecated route name` error. If so, same for similar cases.

PR #5968, opened 2022-08-19
Fix CR10 waypoint on TacStaPkwy

cockroachking:
[https://forum.travelmapping.net/index.php?topic=5111.0](url)

It has been a few weeks and no action has been taken to fix this point yet.


jteresco:
@cockroachking We need to wait until the region's maintainer has a chance to consider the change.  Thanks, @si404 for closing it out for now.

PR #5991, opened 2022-09-03
NOR/NOR

michihdeu:


michihdeu:
@rmihailucian you (and cinx) are affected by the E39 change in Norway

PR #5998, opened 2022-09-05
Cyprus

si404:


michihdeu:
@si404 the HB links for the northern Cyprus routes do not work (no leading zero)

PR #6008, opened 2022-09-09
A few NY forum items

yakra:


rickmastfan67:
@yakra there are no '+' before the depreciated labels on both I-878 & NY-878 files.  Don't know if that will cause any issues.

rickmastfan67:
And on a side note that I just noticed, but shouldn't 'Exit 2' be centered there on the two carriageways instead of just centered on the EB one?  There is a ramp to the WB one from NB JFKExpy. @yakra

yakra:
No issues from missing a leading `+` for AltLabels. They're ignored, same as in a primary label.

The odd geometry is a result of I-878 being EB only, yes.

PR #6020, opened 2022-09-17
IN: delete IN 267 (Plainfield)

mapcat:


Markkos1992:
One more thing @mapcat, the asterisks should be added in the latest KY 237 updates entry as well.

PR #6039, opened 2022-09-25
Update E19, E50, E54, E59

denishanson:


denishanson:
Passed datacheck

michihdeu:
@denishanson jpne is an active system. Changes like E59 opening are newsworthy and require an update entry!

https://travelmapping.net/devel/manual/maintenance.php#newsworthy

PR #6040, opened 2022-09-26
usaca relinquishment changes

ovoss:
Changes discussed on forum.

Datacheck aborted,  Duplicate main list name in ca.ca079bea: 'CA CA79' already points to ca.ca079. Made change, will re-run Datacheck to confirm that worked.


jteresco:
Always appreciate a datacheck problem found and fixed ahead of time! @ovoss

PR #6049, opened 2022-09-29
Added Japan E59 extension to exit 6

denishanson:


denishanson:
Passed datacheck

denishanson:
Passed datacheck

PR #6056, opened 2022-10-03
removing non-existent business loop

the-spui-ninja:


Markkos1992:
The route still needs to be deleted from the usaib files.  I am pretty sure that this will not pass datacheck at the moment.

michihdeu:
yep, this will not work! @the-spui-ninja @jteresco

ovoss:
> The route still needs to be deleted from the usaib files. I am pretty sure that this will not pass datacheck at the moment.

And also from both usaib .csv files, if not done already.

Markkos1992:
I was implying the usaib csv files.  I apologize if I was not clear on that.

jteresco:
I'll remove it from the `.csv` files when I run tonight's update.

the-spui-ninja:
I got it, sorry I missed the first email been running around the hospital for my kid all afternoon

PR #6064, opened 2022-10-07
POL A18

si404:


michihdeu:
The update entries should be `2022-10-07` instead of `2022-10-06`.

I got no notice at the beginning of my log file because I also updated my list file yesterday....

===============

In addition:

> 2 +3 +4 +5 +6 +71 http://www.openstreetmap.org/?lat=51.635280&lon=14.872870

I don't like this approach (like mentioned earlier for similar cases). I traveled the whole DK18/A18 from the German/Polish border to A4 but I got _stolen_ the mileage last night. There is no new alignment but just a new dedication (or signage).

My list file entry is:

> POL DK18 DEU/POL 71

and I'm only credited for the short DK18 now. I'd prefer braking list files in situations like this. Just saying.... 😄

michihdeu:
btw: The short segment has 6 travelers, A18 and DK18Iko have one traveler only (Chris has an E route entry which might credit him for not traveled routes in case of realignments which often happen to E routes), that means 4 more travelers are affected like me. @rmihailucian + emmaf, Aseluit and cougar1989

si404:
> I don't like this approach (like mentioned earlier for similar cases). I traveled the whole DK18/A18 from the German/Polish border to A4 but I got stolen the mileage last night. There is no new alignment but just a new dedication (or signage).

So you'd prefer all the mileage stolen? I'm trying to keep as many people as possible from losing as little mileage as possible, and you want to just deny everyone their mileage until they fix it? You might prefer breaking list files, but your preference on whether people get mileage stolen counts for little as you diligently follow not only the updates table, but the changes to the data, and fix stuff ASAP.

As for which section, 6 people had border-exit 4 and 5 had border-exit 6, so went I with the more people. They will get their mileage back before too long when the DK18 is fully converted to A18, or when the see the update entries and update their .list files.

And, of course, if you had an E36 entry, as Chris has, you'd have lost nothing! Sure, when the E roads move onto new build, you'll gain untravelled mileage, but that's the same with every road - they get updated. DK roads in Poland get changed more so than their concurrent E Roads (because there's many online upgrades to S/A class that don't change the E road). I was hoping more people were in Chris' situation, as that minimises the effect of this change (that I didn't particularly want to do due to the mucking up people, but couldn't see how I couldn't do it).

michihdeu:
> So you'd prefer all the mileage stolen?

Yes. To be triggered by a log file error that I(!) have to check what I really traveled and what not.
Just my 2 cents though....

PR #6085, opened 2022-10-21
BGR A2

si404:


michihdeu:
@si404 2021-10-21 -> 2022-10-21

si404:
Went to the file here on TM/HD to edit that yesterday morning as I was out all day, and when I hit edit it went to si404/HD to edit it there. The whole point was that I had two minutes to fix it, not faff about with lots of extra clicks to pull it in and all that... Grr at Github! I need to remember that it won't let me edit files in the master branch directly now as it thinks it knows best as I keep doing this!

PR #6115, opened 2022-11-02
tasb activation + POL + GA + IDN openings

si404:


michihdeu:
@si404 both CG2.2 update entries refer to espga.cg0022lug
I think one should link to espga.cg0022

PR #6124, opened 2022-11-06
GES/HE/NAQ/OCC/TUR

michihdeu:


michihdeu:
@rmihailucian I've updated two SA FP entries. Two selfref issues are currently open: https://travelmapping.net/devel/datacheck.php?rg=tur (O5 FP, O6 -> D110/D555 as mentioned on the forum)

PR #6147, opened 2022-11-16
BY/HUN/NAQ/ROU/TUR

michihdeu:


michihdeu:
@rmihailucian the turo_con.csv issue had caused a data check error / site update failure. updates.csv entries for removed routes should NOT get HB links and the route name should be the full now. And alt labels for points in use were missing again. I hope that I caught everything....

PR #6195, opened 2022-12-09
F4 / E60 broken concurrency solved

rmihailucian:


michihdeu:
Ah, thanks. I was wrong with the hidden point coords...

btw: your previous updates.csv entry `2022-12-06;Turkey;D965;tur.d270;Route shortened on the east side in Tutak, between actual D965 and former D965 through Tutak` 
indicates the wrong route name.

PR #6208, opened 2022-12-13
MS 7xx-9xx routes

ajfroggie:
Added all remaining and signed MS 7xx-9xx routes.


jteresco:
Temporarily removing MS 741 (empty .wpt) from the CSVs.

https://github.com/TravelMapping/HighwayData/commit/23e2cdc7a903ec64477cd1dd5e75dc2630ce1b4e

PR #6217, opened 2022-12-17
Japan N349, N360, N399 changes

denishanson:


denishanson:
Passed datacheck

denishanson:
Passed datacheck

PR #6222, opened 2022-12-19
Numerous updates to Japan national highways

denishanson:
Mostly related to the national roads being moved to newly open bypasses.


denishanson:
Passed datacheck

PR #6252, opened 2022-12-31
NAQ

michihdeu:
Data verification took 1447.9 seconds with C++
@yakra ;)


yakra:
Ouch!
/me looks at `/home/michih/DataProcessing/siteupdate/cplusplus/logs/siteupdate.log`...
38.5s from finishing reading WPTs to the end sounds about normal.
1294.5s for reading WPTs and 115s for reading systems CSVs says disk access was being real slow for some reason.
That *"some reason"* might be me -- I've had 5 threads of siteupdate running on an infinite loop [looking for crashes](https://github.com/TravelMapping/DataProcessing/issues/549#issuecomment-1368215271) for about 8 hours now. That'll hog a lot of resources and make disk access slower for everyone else. Still though, I'm surprised to see it taking _that_ long. Maybe there were other factors at play, like others running datacheck.sh, DB access, and/or student programs, etc...
I've terminated those processes now. I bet if you ran datacheck again it'd be faster.
Then if you run it again immediately after that, your `hwy_data/` dir will be cached on disk and it will run faster, like you saw [here](https://forum.travelmapping.net/index.php?topic=5338.msg30087#msg30087).

michihdeu:
@yakra Yep, no issue for me. I just wanted you to know. The first run 1-2 hours earlier today took about 5 minutes. I closed PuTTY in-between.

PR #6253, opened 2022-12-31
Christmas time catch-up

si404:


michihdeu:
@si404 Thanks for the update!
- The `DK6 (Rumia)` update entry should have a blank file root.
- Is there a reason why you did not process [Kapiti Expressway extension from Peka Peka to Otaki](https://www.skyscrapercity.com/threads/motorway-openings-around-the-world.1464920/post-181700405) while you've been on NZH? Of course, I understand that you have not yet update other regions you've not been on, but SH1 is odd

si404:
I didn't do the Kapiti Expressway because Wegenwiki is down and that's my go to for motorway openings as much more reliable and helpful than that SSC thread now your thread is gone. The tourist route stuff was triggered by something else. It's a relatively minor change - replacing the point in Otaki on SH1 with two either side and an updates entry about the slight relocation to the east. I'll do it this afternoon.

PR #6315, opened 2023-01-21
jpnh N314-N325 second pass modifications 

denishanson:


denishanson:
Passed cplusplus datacheck

PR #6319, opened 2023-01-22
Truncate MD 331

Duke87ofST:


Markkos1992:
I think this [shield](https://www.google.com/maps/@38.486306,-75.8255601,3a,75y,315.87h,90t/data=!3m6!1e1!3m4!1sGNL5-caFHmw8TUiWqFhPOQ!2e0!7i16384!8i8192) was the only reason I did not suggest truncation myself.  Having said that, I did see that HLR agrees with truncation so truncation it is.

Duke87ofST:
> 

There are also End/Begin state maintenance signs at the intersection with the ramps from US50 EB, which I noted in the field today and were what tipped me off to check HLR.

PR #6381, opened 2023-02-08
cansf preview -> active

mapcat:


yakra:
ping @rickmastfan67

jteresco:
Holding off pending review.

rickmastfan67:
Looks ok to me, sans needing separate news entries for the rename of the DVP & Gardner files for the normal updates area IMO since they were active routes.

mapcat:
@rickmastfan67 Is that really necessary, since I added them as alt names? No one needs to change their .list.

jteresco:
Looks to me like the AltRouteName is sufficient here.

yakra:
Still worth an updates entry IMO.
I've done then for redesignated routes. It's helpful if someone clinches some more mileage on the Creek Turnpike, goes looking for the Creek Turnpike in the HB and can't find it anymore. "Oh, it's OK364 now."

Maybe something as simple as "Moved from Ontario Provincial Freeways to Canada Select Named Freeways".
Might not be worth the bother of mentioning a new .list name or any of that; people can see that in their userlogs.

mapcat:
How's that look now?

rickmastfan67:
Looks good to me now. ;)

PR #6388, opened 2023-02-09
Japan - Remove Byp from filenames

denishanson:
With this changeset Byp has been removed from all Japanese national highway filenames


denishanson:
Passed datacheck

PR #6393, opened 2023-02-11
MS routes, MN 67 reroute, and unused alt label cleanup

ajfroggie:
* Added the remaining Mississippi state routes (mainly the MS 5xx routes)
* Rerouted MN 67 and related changes, adding MN 167 and removing MN 274
* Cleaned up unused alt labels for AL, MN, MS, and VT

PR #6400, opened 2023-02-11
Update systemupdates.csv (usams)

mapcat:


mapcat:
See https://github.com/TravelMapping/HighwayData/pull/6393

PR #6412, opened 2023-02-14
New route ROU DJ229N

rmihailucian:


michihdeu:
@rmihailucian thanks, data check successful.

There is another similar small new road in ROU: https://forum.travelmapping.net/index.php?topic=5401

PR #6420, opened 2023-02-16
MA 8A, 10, 66 point requests

yakra:
https://forum.travelmapping.net/index.php?topic=5407
https://forum.travelmapping.net/index.php?topic=5408


jteresco:
Thanks for these.

PR #6502, opened 2023-03-20
NLD

michihdeu:


michihdeu:
> 

Nah, I learned programming with **https://en.wikipedia.org/wiki/Intel_8085**

PR #6545, opened 2023-04-07
LTU

michihdeu:


michihdeu:
@si404 I've also tweaked some border wps in BLR, LVA and RUS.

michihdeu:
@yakra Data check didn't bother: https://github.com/TravelMapping/HighwayData/pull/6545/files#diff-8bbf5b79c2de0335e5906da485980a74334ed7988bfcbd668f4e2f769c7d24bd

`K197 http://www.openstreetmap.org/?lat=55.616177&lon=22.253189`
`^`

Dunno if it is an issue. Will fix it now.

yakra:
That gets flagged SINGLE_FIELD_LINE then dropped from the DB.

Not a fatal update error, so nothing mentioned in the *`datacheck.sh`* sense of datacheck, but it appears in `datacheck.log`:
```
ltu.k160;^;;;SINGLE_FIELD_LINE;
```

PR #6580, opened 2023-04-21
Hokkaido second pass

denishanson:
Mostly adjustments to waypoint positions
Plus a few roads added
A few roads removed
And several fixes to roads outside of Hokkaido


denishanson:
Passed datacheck

PR #6596, opened 2023-04-29
POL + SRB

si404:


michihdeu:
@si404 
 2023-04-29;Poland;S7 (Janki);pol.s00**8**;Route deleted
->
 2023-04-29;Poland;S7 (Janki);pol.s00**7**;Route deleted

Is the downgrade of DK7 confirmed? Removing from OSM was rejected: https://www.openstreetmap.org/note/3663430

si404:
S7 (Janki) was wholly concurrent with S8 and none of it is now part of the S7 (it was the bit between S2 and DK7 that took the mainline of the '7' route while the missing link that just opened was still being built), so the link is correct. I would add a note, but there's no point as no one was using the route in their list files.

DK7 wasn't reverted on OSM when I made it, and I'm leaving it for now as it 99.999% will be downgraded or renumbered, if it hasn't already been so (the problem is almost certainly that they didn't cover up the signs quick enough for the reverter) - every other DK road gets downgraded (or renumbered) when bypassed by an S road of the same number (with some exceptions, most of which are due to construction of the new route not yet complete, its "one number, one route"). ssc.siskom.waw.pl has removed it as a DK road from its map, and I trust them far more than one person on OSM reverting another.

PR #6598, opened 2023-04-30
GA updates

ntallyn:


Markkos1992:
@ntallyn The updates entry for the GA 32 Truck route says 4-30-2022 instead of 4-30-2023.  That is why it is not at the top of the updates list right now.

ntallyn:
Pull request for the fix submitted just now.

    On Sunday, April 30, 2023 at 10:44:51 PM EDT, Markkos1992 ***@***.***> wrote:  
 
 


@ntallyn The updates entry for the GA 32 Truck route says 4-30-2022 instead of 4-30-2023. That is why it is not at the top of the updates list right now.

PR #6668, opened 2023-06-14
Michih's Balkan Trip planning fixes

si404:


michihdeu:
Oh, it's just the first week out of three ;)

PR #6682, opened 2023-06-27
DenaliParkRd: Add Visible wpt for EastForkBridge

ssoworld:
Using Shaping point X348051 - closure here due to pretty rocks landslide until 2026 busses do not go past it. see https://forum.travelmapping.net/index.php?topic=5551


ssoworld:
@Duke87ofST @ovoss please review

Duke87ofST:
Keeping X348051 as an alt label isn't necessary.

I'd move the point to be at the bridge: http://www.openstreetmap.org/?lat=63.556427&lon=-149.742365

ssoworld:
shaping point removed - the waypoint doesn't go at the bridge. using http://www.openstreetmap.org/?lat=63.559108&lon=-149.794271 instead

PR #6700, opened 2023-07-08
Add US 12 Touchet bypass

compdude:
Adding newly-opened Touchet bypass for US 12 

https://forum.travelmapping.net/index.php?topic=5431.0


jteresco:
This changed a lot in `wa.us012.wpt` but it appears that most of the changes were removing extraneous spaces at the end of lines.  This seems like a good change, worth the large diff.

compdude:
Yeah, I copy-pasted the whole wpt file into the waypoints editor, made my changes (including some other minor point relocations), and then copy-pasted everything back into the waypoint file.

PR #6717, opened 2023-07-23
KY & MI

mapcat:
KY: KY 1560 truncated
MI: NMP


rickmastfan67:
@mapcat KY-22's mulitplex with US-25 here would be broken if pulled in.  See forum post.

PR #6725, opened 2023-07-27
ESP, IDN, POL

si404:


michihdeu:
@si404 
`2023-07-28;(Spain) Basque Country;BI625;esppv.bi625;Route truncated at ? end from BI627 to N634 (former eastern junction with N634)`

I guess that `?` should be `north`

PR #6756, opened 2023-08-11
Update systems.csv

neroute2:
not sure why mexsf is still in devel...should be ready for preview


michihdeu:
@neroute2 please update systemupdates.csv too

neroute2:
> @neroute2 please update systemupdates.csv too

Done

PR #6770, opened 2023-08-17
forum issues (ITA+AUS+Asia)

si404:


michihdeu:
@si404 an updates.csv link is broken:
`ita.viamarpol`
->
`ita.viamarpolo`

PR #6804, opened 2023-09-07
Brasil

neroute2:


rickmastfan67:
Word of advice @neroute2, please make all the changes first, then create the Pull Request.  I got nailed with over 300 emails today.

neroute2:
> Word of advice @neroute2, please make all the changes first, then create the Pull Request. I got nailed with over 300 emails today.

Didn't realize that was a thing. I've been creating a pull request and then editing it as my modus operandi, but I guess that has unforseen consequences, so I will change my process.

rickmastfan67:
Yeah, Github sends an email for each change after you submit a pull request.  So, if you change the name of 300 files in it, that's 300 separate emails.  You don't get them, since you created the pull request, but others that are auto-subscribed for when new PR are submitted, get nailed by it.

rickmastfan67:
Here's a quick example of all the emails I got today, just from this pull request sadly due to the '1 change = 1 email' thing that GitHub does.


example

jteresco:
Thanks for this great expansion.

Looks like the site update went through but there are lots of errors because there is no region BRA-SP defined in regions.csv . Could you add the Brazil regions to that file? @neroute2

neroute2:
> Looks like the site update went through but there are lots of errors because there is no region BRA-SP defined in regions.csv . Could you add the Brazil regions to that file? @neroute2

I think I've done this properly now.

michihdeu:
> > Word of advice @neroute2, please make all the changes first, then create the Pull Request. I got nailed with over 300 emails today.
> 
> Didn't realize that was a thing. I've been creating a pull request and then editing it as my modus operandi, but I guess that has unforseen consequences, so I will change my process.

I don't get emails but notifications in the webbrowser. That's annoying too. I just didn't dare to mention it in the last days.
On the contrary, I don't get any notification from @si404 nor @jteresco because they directly commit to master. That's annoying too, e.g. in case of railway data when I would edit the same file like si404 did. It just has not yet happened because we do not yet edit updates.csv nor datacheckfps.csv etc.

PR #6824, opened 2023-09-17
KS 14/96 relocation & Beatrice NE truck routes

yakra:
https://forum.travelmapping.net/index.php?topic=5211


yakra:
Needs updates entries. Will add along with [KS14/96 relocation](https://forum.travelmapping.net/index.php?topic=5537).

PR #6876, opened 2023-10-11
tasc review and A38M exit numbers from forum

si404:


hwy_data/ENG/gbna/eng.a6019.wpt
```diff
@@ -1,2 +1 @@
-A453 http://www.openstreetmap.org/?lat=52.945380&lon=-1.152964
-A60 http://www.openstreetmap.org/?lat=52.946634&lon=-1.140164
+
```
rickmastfan67, line 1: Would probably help if you deleted the file, instead of just leaving it as a 1-byte. ;)

PR #6882, opened 2023-10-14
Poland A18 + S61, Latvia A7, Serbia A8 openings

si404:


michihdeu:
@si404 you missed the csv file entries for Serbia

PR #6974, opened 2023-11-12
NY forum fixes

yakra:
https://forum.travelmapping.net/index.php?topic=5857.msg33026#msg33026
`+PigHillRd` is retained as an AltLabel, only in use by @jteresco, in case you feel like updating to the primary label.

---

https://forum.travelmapping.net/index.php?topic=5889
https://forum.travelmapping.net/index.php?topic=5890


jteresco:
Updating my list to remove the old label.

PR #7136, opened 2024-01-05
Apparently when you make a new patch it takes all the stuff from the old one; Colombia is ready for preview!

neroute2:


michihdeu:
@neroute2 Did you run data check?

I've sync an hour ago and get error:
`1: colsf.csv: root col.anivialcre not matched by any connected route root.`

michihdeu:
In regards to your PR title: I wonder why you were able to open a new PR. When I commit changes with GitHub Desktop, they are always put into the only opened PR. I can only submit a new PR when no PR is opened for the repo. I guess that the web frontend uses different commands behind the scenes?

jteresco:
Successful site update with https://github.com/TravelMapping/HighwayData/commit/77ed17190eee9e0e00190492a66bf7a27ed97efc in place.

PR #7153, opened 2024-01-10
Highway63 email

si404:


hwy_data/_systems/usamo.csv
```diff
@@ -1,4 +1,4 @@
-systemName;region;route;banner;abbrev;city;root;altRouteNames
```
rickmastfan67, line 1: Wouldn't removal of this line also break usamo.csv?

hwy_data/MO/usamo/mo.mo094.wpt
```diff
@@ -34,8 +34,8 @@ MOsDD_StC http://www.openstreetmap.org/?lat=38.652824&lon=-90.776424
 MOsD_StC http://www.openstreetmap.org/?lat=38.703363&lon=-90.709820
 I-64 http://www.openstreetmap.org/?lat=38.711735&lon=-90.690336
 WolRd http://www.openstreetmap.org/?lat=38.727706&lon=-90.644245
-8A(364) +MidRivDr +MO364_W http://www.openstreetmap.org/?lat=38.737883&lon=-90.633302
-8B(364) http://www.openstreetmap.org/?lat=38.740444&lon=-90.617809
+8 +8A(364) +MidRivDr http://www.openstreetmap.org/?lat=38.737883&lon=-90.633302
```
rickmastfan67, line 37: Should be '8(364) +8A(364) +MidRivDr'.

hwy_data/_systems/usamo.csv
jteresco: I don't think it will break as long as there's something on that first line, but I would prefer to keep the CSV-standard line at the top, as it makes it easier to auto-import our CSV files into other tools and have it build data structures appropriately.  I'd like to put it back as it was.

PR #7170, opened 2024-01-19
NC 57: Should be true fix for US158_W

Markkos1992:
Clarify that mapmikey has responded on forum before merging.

https://forum.travelmapping.net/index.php?topic=6042.msg33786#msg33786


Markkos1992:
mapmikey has responded on the forum.

PR #7195, opened 2024-01-30
Sapulpa 2024-01-29

yakra:
https://forum.travelmapping.net/index.php?topic=5496
**OK33:** East end truncated from US 75 Alt and OK 97 to OK 117.
**OK66:** Removed from Dewey Ave and relocated onto OK 117 & OK 117A (Mission St) between waypoints DewAve_W & DewAve_E.

Ping @dietrichmd @bejacob _@mapcat @si404_ @bhoward @ozarkman417 _@theradison @scott5114_ @ua747sp
...and a few others. Something's changed at GitHub and now my script won't pull usernames 100% of the time. :(

### DIFF
--- a/hwy_data/OK/usaok/ok.ok033.wpt
+++ b/hwy_data/OK/usaok/ok.ok033.wpt
@@ -63,5 +63,4 @@ OK48 http://www.openstreetmap.org/?lat=36.003146&lon=-96.387091
 N3750 http://www.openstreetmap.org/?lat=35.988355&lon=-96.335335
 I-44 http://www.openstreetmap.org/?lat=35.987313&lon=-96.193972
 OK66_W http://www.openstreetmap.org/?lat=35.985646&lon=-96.191182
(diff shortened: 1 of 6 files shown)


dietrichmd:
so explain that to me like i'm 5...does this mean I need to change my submission method?

rickmastfan67:
> so explain that to me like i'm 5...does this mean I need to change my submission method?

It means labels changed that affect your list file for those routes mentioned above.  You should double check to make sure your mapping is still accurate.

ua747sp:
old routing was in red, so i'm in the process of removing my travels
between the two points circled in blue. based on the forum, looks like
there might be a historic US66 route drawn between those two points along
the red, so will check back periodically until that's added.

[image: image.png]

rickmastfan67:
@ua747sp FYI, images don't work as attachments if you send via email.  You need to post them via Github itself.

ua747sp:
Thanks. Let me try again here.

image

PR #7203, opened 2024-02-08
Venezuela is as ready as it will be

neroute2:


jteresco:
One error, I will switch to "Jua" from "Juan" and rerun.

<pre>
[64.7] ABORTING due to 1 errors:
1: Abbrev > 3 bytes in vena.csv line: vena;VEN;AutGranMar;;Juan;Autopista Gran Mariscal de Ayacucho (San Juan);ven.autgranmarjuan;
</pre>

PR #7211, opened 2024-02-12
MAR N23

rmihailucian:


michihdeu:
@rmihailucian I took the opportunity and added wps for the old route south of Demnate

PR #7291, opened 2024-03-26
Brasil error fix

neroute2:


neroute2:
Yeah, I have no idea how to run datacheck. I deliberately introduced an error into the Brasil2 branch and made that one default, then ran datacheck in putty and it gave no errors.

michihdeu:
> Yeah, I have no idea how to run datacheck. I deliberately introduced an error into the Brasil2 branch and made that one default, then ran datacheck in putty and it gave no errors.

First, there is no _neroute2_ folder in the _fast/home_ directory.

jteresco:
His is in `/home/neroute2` (can always access a user's home directory by `~username`).

I suspect the problem might be that the branch to be checked isn't pulled in from GitHub.

jteresco:
@michihdeu thanks for assisting, this is a different method than the one I use as a repository owner.

michihdeu:
> I suspect the problem might be that the branch to be checked isn't pulled in from GitHub.

It's from 2023-03-13

michihdeu:
@neroute2 I think you need to run _git clone https://github.com/neroute2/HighwayData.git_ before running the datacheck command itself.

(I always run that when I used the method you need to use)

neroute2:
> @neroute2 I think you need to run `git clone https://github.com/neroute2/HighwayData.git` before running the datacheck command itself.
> 
> (I always run that when I used the method you need to use)

fatal: destination path 'HighwayData' already exists and is not an empty directory.

jteresco:
The clone step is only for the initial setup.  After that a `git pull` should update the clone to match what's up on GitHub.

neroute2:
> The clone step is only for the initial setup. After that a `git pull` should update the clone to match what's up on GitHub.

fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).

PR #7293, opened 2024-03-27
Brasil

neroute2:
adding bradf; if I could get datacheck working I'd run it


jteresco:
Let's try this:

```
cd ~/travelmapping/HighwayData
git pull
git checkout Brasil
/fast/tm/datacheck
```

neroute2:
> Let's try this:
> 
> ```
> cd ~/travelmapping/HighwayData
> git pull
> git checkout Brasil
> /fast/tm/datacheck
> ```

The first line works without the travelmapping directory, and the others work too. Datacheck then runs successfully. I will try later with a deliberate error, but it looks like this may be it.

jteresco:
Right - I keep my repository clones in a travelmapping directory, but the instructions on this don't indicate to do that.

Hope this is what we need to get you running.

neroute2:
YESSS it works!
https://github.com/neroute2/HighwayData/commit/d720f8a7292139cb79b91e64e8d3ecfce092ec12
[17.1] ABORTING due to 3 errors:
1: Could not find Route matching ConnectedRoute root braac.br317 in system brabr.
2: [Errno 2] No such file or directory: '/home/neroute2/HighwayData/data/BRA-AC/brabr/braac.br317hahaha.wpt'
3: brabr.csv: root braac.br317hahaha not matched by any connected route root.

jteresco:
Good!  I think the usual directions and the `datacheck` script assume the repository already has the branch you want to check checked out.  I could probably add an optional parameter to `/fast/tm/datacheck` that would do the `git checkout` step.

https://github.com/TravelMapping/DataProcessing/issues/618

PR #7329, opened 2024-04-12
More PA National Road and Lincoln Hwy Updates

Markkos1992:
@rickmastfan67 I went ahead and did the WV revisions by the PA Line on US 40 and the National Road.

https://forum.travelmapping.net/index.php?topic=6158.msg34299#msg34299


rickmastfan67:
@Markkos1992 ty for pinging me.  US-40 updated in my 'local' files so the point doesn't get accidentally deleted in the future.

PR #7371, opened 2024-05-02
extension of M1610

neroute2:
only changes to existing files (compared to the Mexico2 pull request) so no need to run datacheck


neroute2:
These three pulls should be run from oldest to newest so each only has the files changed for it and not the ones for the previous pulls.

michihdeu:
@neroute2 I think that some 30% of your links are broken. Please check.
e.g.
https://travelmapping.net/hb/showroute.php?r=mexgto.mex110d
https://travelmapping.net/hb/showroute.php?r=mexemex.emex007d

unfortunately, there is no automated check yet: https://github.com/TravelMapping/DataProcessing/issues/62

neroute2:
> @neroute2 I think that some 30% of your links are broken. Please check. e.g. https://travelmapping.net/hb/showroute.php?r=mexgto.mex110d https://travelmapping.net/hb/showroute.php?r=mexemex.emex007d
> 
> unfortunately, there is no automated check yet: [TravelMapping/DataProcessing#62](https://github.com/TravelMapping/DataProcessing/issues/62)

Both of those are deleted (renamed) routes...

michihdeu:
@neroute2 If the route was renamed, it should link to the new route. If it was fully removed from HB, the field should be empty.

neroute2:
That makes sense. Will fix.

michihdeu:
@neroute2 One of the 2024-04-30 update entry link is still broken: mexbcs.mex0019d

PR #7374, opened 2024-05-03
Mexico updates on the Atlantic and Pacific coasts (including major new pieces of 180D)

neroute2:
data check successful


neroute2:
Data check successful again after merging Mexico2. I'm trying different ways of doing multiple updates that may conflict (due to changes to mexd.csv and additions to updates.csv) and this seems better than multiple pulls.

PR #7408, opened 2024-05-19
usatr / srbr

si404:


michihdeu:
@si404 thanks for adding srbr system. You forgot the systemupdate.csv entry though.

PR #7432, opened 2024-05-31
I-11 extended, I-515 deleted, waypoint tweaks around Las Vegas

cl994:
- NV I-11 extended north to NV 157 over US 95
- I-515 deleted, replaced with extended I-11. I-515 retained as alternate label for I-11.
- I-515, US 93, and US 95 waypoints changed to reflect that I-11 is the primary route
- Intersecting route waypoints changed to reflect that I-11 is the primary route
- Minor waypoint tweaks around Las Vegas to reflect current endpoints and remove references to decommissioned routes

This should go through without breaking anything, but please let me know if you see any concerns in the code.


rickmastfan67:
Aren't exit numbers changing too?  Just curious.

cl994:
Nope, no exit number changes yet. I assume that will happen eventually, but it took a few years with 580 as well.

rickmastfan67:
Alright.  Just wanted to be sure.

PR #7469, opened 2024-06-23
NH & MA relocations

yakra:
https://www.dot.nh.gov/projects-plans-and-programs/project-center/dummer-cambridge-errol-16304


yakra:
Fall River Mass update incoming.

yakra:
Makes me a little sad to push this Fall River update.
The old configuration had some fine examples of TM weirdness.
* Concurrencies in one direction but not the other, with hidden points to break them
* Counterintuitive labels at the E end of the Veterans Bridge
* 3 overlapping single-route traces between US6_E and Brightman St
* Last but not least, I developed a CLOSED_OPEN_COLOC datacheck back in... holy crap, 2020, but just stashed it without ever committing or reporting any flagged errors in the forum. `BriSt` would have been a rare example of a false positive, with an exit from the US6/MA138 frontage road, but no connection to the MA79 mainlanes.

Ready to rock.

PR #7470, opened 2024-06-23
US166/400 freeway

yakra:
https://www.koamnewsnow.com/news/kansas-news/now-open-kdot-opens-four-lane-expressway-in-cherokee-county/article_e933c86e-2f40-11ef-bfaf-6fceb69cc03e.html

### DIFF
--- a/hwy_data/KS/usaks/ks.ks026.wpt
+++ b/hwy_data/KS/usaks/ks.ks026.wpt
@@ -1,4 +1,5 @@
-US166/400 http://www.openstreetmap.org/?lat=37.023636&lon=-94.642378
+US166 http://www.openstreetmap.org/?lat=37.020531&lon=-94.642472
+BagRd +US166/400 http://www.openstreetmap.org/?lat=37.023636&lon=-94.642378
 DenDr http://www.openstreetmap.org/?lat=37.043437&lon=-94.643118
 21stSt http://www.openstreetmap.org/?lat=37.060353&lon=-94.639213
 KS66 http://www.openstreetmap.org/?lat=37.074932&lon=-94.639192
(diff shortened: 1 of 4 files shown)


hwy_data/KS/usaks/ks.ks026.wpt
```diff
@@ -1,4 +1,5 @@
-US166/400 http://www.openstreetmap.org/?lat=37.023636&lon=-94.642378
+US166 http://www.openstreetmap.org/?lat=37.020531&lon=-94.642472
```
rickmastfan67, line 1: Shouldn't the 'new' end point label be US400/166 instead? ;)  I know 'US166/400' is hidden under the old end, but still, doesn't prevent a flip of the routes themselves in the 'new' label. for the new end :) @yakra

yakra: This was done intentionally to avoid breaking .list files.

yakra: Having 2 routes mentioned in a label is allowed but not required.

rickmastfan67: > This was done intentionally to avoid breaking .list files.

I know that.  But there wouldn't have been any issue using 'US400/166', as that wouldn't have been the same label. ;)

> Having 2 routes mentioned in a label is allowed but not required.

True. But with two major US highways there, I'd assume that they'd be both mentioned if possible.  I mean, it's not like US-166 was paired with a state highway there.  If that was the case, I'd fully agree with dropping the state number from the label.

yakra:
Yeah no. I'm not gonna slash a higher numbered route before a lower numbered route (of the same system) in a label.
It'll be grand.

hwy_data/KS/usaks/ks.ks026.wpt
yakra: Yeah no. I'm not gonna slash a higher numbered route before a lower numbered route (of the same system) in a label.
It'll be grand.

PR #7481, opened 2024-06-28
Update bc.bc099.wpt

ovoss:


yakra:
The first commit in #7482 has the same deltas, but a different parent.
From there's there's another child commit further changing one of the new labels.
Closing this commit as it'd just produce a merge conflict for no gain.

rickmastfan67:
Safe to destroy this branch @ovoss?

ovoss:
Yes. I've removed that and another stray branch.

PR #7485, opened 2024-06-29
ENG + ESP-GA

si404:


michihdeu:
@si404 SC-30's update link is broken. Leading zero missing.

PR #7525, opened 2024-07-07
CHN GD-G78 adjust junctions

eeffaann:

PR #7543, opened 2024-07-12
Merge various disconnected CHN-G routes

eeffaann:


eeffaann:
!!! DATA CHECK SUCCESS !!!

yakra:
7914ef05bf4a54103a02d0b6f395b8cf53847fcb appears to be a squash merge rather than a merge commit

jteresco:
Going to hold off until I have a better understanding of @yakra's concern and we know why @eeffaann did this the way it was done.

eeffaann:
OK I rebased my fork and the commits should be cleaner. The previous version basically looked like I am re-checking changes others made in the week or so I had this branch un-synced and rebased.

yakra:
Cloned; looking things over in gitkraken

yakra:
I'll try to type something up this evening.

yakra (commented):
6f4071e0de0acd96dc81e470d5fca3b70294fc1b:
  chnsc.sa2.wpt may want to have a different name, or to be in a different system?
  Or maybe it's waiting around to be part of another system to be drafted at some point in the future?

---

### Overall:
I still see some extraneous merges (squash merges specifically), though they're not something that'll mess up the history for anyone else's files.

The first commit in this PR is be11798, making the ancestor, before everything here gets started, d672345.

In `eeffaann:CHN-merges`, we start off with 8 commits affecting only Chinese files. So far so good.
Meanwhile, pull requests are merged into `TravelMapping:master`.
Then, @jteresco commits 5dfa1f6 a "Fix for #7525 problem."
After that, 6 more commits on `CHN-merges` affecting only China.
Still, so far so good.

Then, there's ee9fd2b.
This is a child of 5dfa1f6, [merging the 14 commits](https://github.com/eeffaann/HighwayData/pull/40) into `CHN-merges`.
Running the commands `git diff d672345 300d82e` and `git show ee9fd2b`, we can see the DIFFs are the same.
On its own, there's nothing wrong with this; it can be a good way to keep a cleaner more concise commit history. If done properly.
A good rule of thumb is to only do this ~on branches~ with commits that have not been pushed to remote yet. That was my concern with 7914ef0 upthread, which caught other people's files up in the net.
Again, on its own, nothing wrong here per se. Still, SFSG.

Development could continue on either branch at this point. One might assume this one due to the squash merge here.
It's from here out that things get a bit weird.

2 more commits on `CHN-merges`, fdd2e19 & aec7d3f.

Then, 7f16a1c, which merges the pre-rebase `eeffaann:master` into `CHN-merges`.
This introduces the upstream commits, other people's pull requests that got merged. Nothing inherently good or bad about this other than that it introduces another commit to the history. While keeping our own forks of `HighwayData` relatively up-to-date with `TravelMapping:master` is probably good practice overall, it's not crucial most of the time. We only really need to watch out for files that are likely to be edited by multiple people, such as systems.csv, systemupdates.csv, or CSVs for systems with multiple contributors such as usaus.csv or usai_con.csv.
But there's one more commit here in the history that gets merged: ee9fd2b, which if we recall, duplicates the diffs from the 14 prior commits in `CHN-merges`. Meaning, all these diffs are stored in the commit history **twice**.
We have twice the deltas, wasting a bit of space, but this may not amount to much if git is able to recognize & deal with the duplication and compress it effectively when compressing the objects into packs.
Sure, it makes the commit history harder to follow, but... who ever tries to follow the HighwayData commit history? :D HighwayData's commit history is, in a word, filthy :rofl:, with dozens of contributors, thousands upon thousands of commits, and some occasionally tangled spaghetti of many many branches. *(And some... other stuff.)* That's just how it's gonna be.

---

After that, a little more of the same:
* cddef69 is [another squash merge](https://github.com/eeffaann/HighwayData/pull/41).
  Git thought it was merging 16 commits this time, as none of these exact commits* existed in the commit history of this branch.
  However, the first 14 of them got squashed into ee9fd2b, so the only diff here (on the pre-rebase `eeffaann:master` branch) is the same as the last two (fdd2e19 & aec7d3f) squashed.
* 1 more commit on `CHN-merges`, 540f6cf.
* c1f4aea, another merge commit.
  This only merges in cddef69, the one commit on pre-rebase `eeffaann:master` not yet included in the history of `CHN-merges`.
  However, all of the consituent commits that were squashed down to form that commit *are* included.
  Thus, this commit has no diff at all from its first parent.
* 6 more commits on `CHN-merges`.
* `eeffaann:master` reset to 459f497. Most likely this was `TravelMapping:master` at the time.
* 1 more [merge](https://github.com/eeffaann/HighwayData/pull/47) commit, merging `CHN-merges` into `eeffaann:master`.
  Useful for those who [want to run datacheck on noreaster and aren't comfortable checking out branches other than master](https://forum.travelmapping.net/index.php?topic=6351).
  Other than that, this just merges the upstream commits with the CHN topic branch. Not super crucial as noted above; another way of doing that would be to just merge into TM:master from `CHN-merges`, then FFWD `eeffaann:master` afterward.

---

# TLDR
It's a bit weird, yeah, but **by no means the end of the world.**

Just consider this a long-winded breakdown of what happened, and please don't think I say all of that to advocate for the perfect to be the enemy of the good.
If we required commit history to be absolutely pristine every step of the way, and for everyone contributing to have MAD GIT SKILLZ, that's just end up being a significant barrier to participation, and a lot less would get done.

Jim, if you're OK merging this in, go for it!

eeffaann:
> [6f4071e](https://github.com/TravelMapping/HighwayData/commit/6f4071e0de0acd96dc81e470d5fca3b70294fc1b): chnsc.sa2.wpt may want to have a different name, or to be in a different system? Or maybe it's waiting around to be part of another system to be drafted at some point in the future?

Any file renamed to Sxxxx means that the freeway was just demoted to the provincial systems. Better than deleting it even though those systems don't exist yet I suppose.

PR #7596, opened 2024-07-30
ME9 & ME15Bus point density, coords & labels

yakra:


yakra:
I'll be opening another PR to add [AR600](https://forum.travelmapping.net/index.php?topic=6327) in a few minutes. Just running datacheck on BiggaTomato.

PR #7604, opened 2024-08-01
HUN F260, F222 and ROU DN76B

rmihailucian:
new national roads in HUN and ROU


jteresco:
Looks like we forgot datacheck:

<pre>
[5.2] ABORTING due to 1 errors:
1: [Errno 2] No such file or directory: '/fast/home/terescoj/travelmapping/HighwayData/data/ROU/roudn/rou.dncora.wpt'
Thu Aug  1 22:40:31 EDT 2024
</pre>

jteresco:
Removing that route from the csvs and trying again.

jteresco:
Looks like it's working.  The csv file entries should be restored when the dncora wpt file is added.

si404:
The route was renamed, so the .csv line doesn't need readding, save maybe as an AltRouteName for DNVO1Z. There was no update entry, and I would have thought it would need that. And while that's being adding, the typo in the below line should be fixed.

> 2024-08-01;Romania;DN**6**6B;rou.dn76b;New Route

rmihailucian:
Did the fixes, should come with the next update.

PR #7621, opened 2024-08-08
CYP A22 & POL A2 openings

si404:


michihdeu:
@si404 DK92Sie is missing in csv files

PR #7648, opened 2024-08-21
POL cleanup and prep for country split system

si404:


systems.csv
michihdeu, line 1: @yakra do we need a new system category e.g. `deprecated`?

michihdeu: @si404 I said no......

jteresco:
Looks to me like there's some discussion remaining here, so I am delaying this merge for tonight.  @si404 @michihdeu

PR #7649, opened 2024-08-21
I-70/I-76, PA 120, and PA 655 Point Revisions

Markkos1992:
https://forum.travelmapping.net/index.php?topic=6400.0

### DIFF
--- a/hwy_data/PA/usai/pa.i070.wpt
+++ b/hwy_data/PA/usai/pa.i070.wpt
@@ -64,7 +64,10 @@ WV/PA +0 http://www.openstreetmap.org/?lat=40.095736&lon=-80.519228
 +X8D http://www.openstreetmap.org/?lat=40.090384&lon=-79.201984
 +X581764 http://www.openstreetmap.org/?lat=40.076875&lon=-79.179786
 +X738802 http://www.openstreetmap.org/?lat=40.075707&lon=-79.165077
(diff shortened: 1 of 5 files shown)


yakra:
https://travelmapping.net/devel/manual/wayptlabels.php#multipledesignations

yakra:
> because it would take hours for me just to debate what the

LOL I know what you mean.
Today I came across an area in Connecticut that could use some coord tweaks & potential new points to clean up the mapview.
After considering doing it this way, that way, the other way, graph connections, the manual, and what *is and is not* 1PPI, I said sod it all and went no-build other than exit numbers.

For cases like this I usually gravitate toward calling the route itself primary
* `OldUS202_Aub` on US202
* `ME4_Aub` on ME4
* `ME11_Aub` on ME11
* `ME100_Aub` on ME100

Markkos1992:
Yeah, if this was under PennDOT jurisdiction, then it would be SR 0070 meaning that I should be using I-70 labels (which makes sense for the I-70/I-79 concurrency in Washington).  However, the PTC jurisdiction makes this very tricky.  I could go with I-76 labels based on the SR 7076 designation PennDOT gives it, but I have no clue if the PTC even recognizes that.

rickmastfan67:
Well, IMO, since the exit numbers match I-76 mileage, that's the primary route.  That or MM's should be what decides stuff like this IMO, if they're posted that is of course.

https://maps.app.goo.gl/4hMikLiGST48ZLsf9

PR #7652, opened 2024-08-22
Remove N169Dor and N169Tad

denishanson:
Old roads reverted to local control


denishanson:
Passed datacheck

denishanson:
Passed datacheck after N417 changes

PR #7708, opened 2024-09-12
Delete hwy_data/SC/usansf directory

rickmastfan67:
Delete old copies of SC routes that had been in usansf, due to them being in usasc already & being activated there. Thus, making these old files unnecessary. This removes them from the 'unprocessedwpts.log' file.


rickmastfan67:
@mapmikey

PR #7784, opened 2024-10-09
30 Crossing

yakra:


yakra:
Ugh wait, let's hold off on this again...

yakra:
Okay yeah. Ready to rock.
At some point this may deserve another look north of the river, but AFAICS construction isn't that far along yet.
All set for now.

PR #7786, opened 2024-10-10
AUT/CZE/PAC/SWE

michihdeu:
Couldn't run data check https://github.com/TravelMapping/DataProcessing/issues/653#issuecomment-2405673383
I changed many csv files. Thus, it's possible that there are errors. Be careful or simply don't merge it today @jteresco


michihdeu:
Data check successful. Merging.

PR #7803, opened 2024-10-17
Update chngd.ah001.wpt

eeffaann:
Fix per https://forum.travelmapping.net/index.php?topic=6527.0;topicseen


yakra:
> Suggest G4/G1508 as the label.
https://travelmapping.net/devel/manual/wayptlabels.php#bothroutes

jteresco:
Will await review before merging.

michihdeu:
> > Suggest G4/G1508 as the label.
> > https://travelmapping.net/devel/manual/wayptlabels.php#bothroutes

@yakra what do you mean? I've requested a review from you the other day to clarify it.

yakra:
Sorry, thought my comment in itself was adequate.
`17(G1508)` was deleted, which is good.
`2150(G4)` still exists, but with the same label as before. Suggest changing it to `G4/G1508` as per the manual.

jteresco:
Thanks, I will then wait for that suggested change to come in before merging.

eeffaann:
Updated

PR #7809, opened 2024-10-18
corrections or ITA Split

rmihailucian:
@yakra please run it again, let's see if all errors were corrected.


michihdeu:
@rmihailucian I'd like to know if it is compatible with #7810 @si404 please review!

si404:
The only file we've both changed is itaa2_con.csv, but we did different lines, so it should be OK. I'll merge this in, and fix any conflict with my branch if some are created.

yakra:
Keep in mind, when two branches both datachecks clean, merging them together is virtually guaranteed to datacheck clean too.

When we run /fast/tm/datacheck on noreaster, what it tests out is our own personal fork of the repo, the purpose being to make sure it's clean before merging to master. The assumption is, if master is known clean, and a contributor's fork is clean, merging them will be all good.

If we had to worry about doing a datacheck *after* merging, well... that'd be a bit of a problem wouldn't it? :)

michihdeu:
> If we had to worry about doing a datacheck _after_ merging, well... that'd be a bit of a problem wouldn't it? :)

I don't get it. Simply sync your fork again and run /fast/tm/datacheck.

yakra:
> The assumption is, if master is known clean, and a contributor's fork is clean, merging them will be all good.

I wasn't *quite* right there:
Just looked through every instance of `add_error` in the codebase to be able to say this confidently. The only exception I could come up with was system .csv & _con.csv files. Suppose that...
* Branch 1 adds a root or .list name. Merges in clean. Datachecks clean.
* Branch 2 adds the same root or .list name on another line in the same file. Merges in clean. Datachecks clean.
* Merging both of these together can result in a duplicate root or .list name, and an `ErrorList` error.

Most likely, this would involve creating a new route file, which would introduce a merge conflict.
Other than that, if 2 contribs are both just adding some CSV lines to take care of an *unprocessedwpt*... yeah.
Clean merge & datacheck on 1 by itself, clean merge & datacheck on 2 by itself, merging them *together* is clean, but datacheck fails.

Probably though, 2 contributors won't be working on a single region. (Unless, typo? "MA" for "ME"?)
If it's just one person, he'll probably be on top of what he's doing enough to not duplicate his work like that.

That's the only exception I can think of; I do believe other than that clean merge = clean datacheck.

---

Other than that?
* The majority of `ErrorList` errors involve malformed single lines in some CSV or other.
  In these cases clean merge -> clean datacheck.
* The few remaining involve missing files. These should be OK as long as the original (pre-merge) branches datachecked clean.
  To avoid an error, every CSV referring to a file will refer to a file that is in the repo. As long as the merge is clean, the files won't go away. If there are merge conflicts, just gotta be sure they're resolved correctly.

si404:
Trying to decipher the confusing techobabble from @yakra  - do you think I ran datacheck to check if there was no conflict between the two pulls? I did not!

Running datacheck (with the itaa2 and itass2 systems unchecked) was about checking that the changes of this pull did what they were meant to do (fix the errors in those .csvs) and wouldn't break the site. This was specifically requested by @rmihailucian be done, and I did it as I might as well do it while addressing @michihdeu 's query, given checking there were no conflicts meant that I had this pull merged into my branch anyway, and so could datacheck it very easily.

PR #7810, opened 2024-10-18
Italy systems clean up

si404:
@yakra - hopefully this is fine for you. I've split eure2 into its two functions (bihe and itae). I've renamed eursf to itasf. I've moved all the split-related systems to the bottom of systems.csv for easy access.

(this pull also moves RA16, but does nothing else)


michihdeu:
@si404 is it compatible with #7809 ?

si404:
Looks like no conflicts with #7809 pulled in.

yakra:
Will review later on.
In the immediate term, strongly suggest *not* merging this just yet.

For now, working on finalizing https://github.com/TravelMapping/DataProcessing/issues/492, which will become important later as we continue validating data for the ITA split.

yakra:
Looking into this now.
There's some technical stuff with `siteupdate --splitregion` I'm trying to understand & get sorted.

yakra:
I've opened a PR for an alternative to this at #7821.

PR #7821, opened 2024-10-19
#7810 redo

yakra:
@si404, like you said - "hopefully this is fine for you." :)

Sorry to be TLDR yet again, but here are my notes on why I'm doing what I'm doing...

Proposing this as an alternative to #8710 because the system renames there break `siteupdate --splitregion`, due to the way it finds the systems it needs to pay attention to.
I was faced with a choice between rewriting the splitregion routines, and God-knows-what kind of debugging that would entail, and just redoing #7810. I chose the latter. Just so much simpler.

Going no-build except for the meaningful hwy data changes, i.e. the RA16 change, the E74 split, and creating bihe.
The logic here is, this stuff will all have to change anyway when it's go-time:
• Split systems will be "dismantled" and have their routes spliced into their parent systems.
• The new `hidden` systems will be created from the vacated chopped routes from ITA (not its subregions). ConnectedRoutes will chave one chopped Route each.
• systems.csv entries.
Why not just hold off on making unnecessary changes till it's time to make the necessary ones.

Some notes on changes to **systems.csv** I didn't carry over from #7810:
* `itasf` was coloured `TMcyan`, which isn't defined. Kept the original line, with `eursf2` coloured `TMteal`.
* This *wasn't* changed in #7810, but should `zafrtr` have `ZAF` for its country instead of `AFR`?
* I didn't keep the active->preview changes for the ZAF systems. Won't matter if these systems are and will remain disabled.
* Naming:
  The "split hidden system" nomenclature concerns me. Sure, we can call it what we want, siteupdate won't be bothered, but...
  It suggests a misunderstanding of what a hidden system will be.
  These are only "hidden" in the sense that they're *disabled*, commented out.
  A real `hidden` system (which will be their `Level` as opposed to active/preview/devel) won't be a split (among the new smaller regions) system, but have the chopped routes (covering all of ITA, unsplit) taken from the parent systems.
  Existing names are clearer & less misleading IMO.
  (I suggest saying "disabled systems" to refer to commented-out ones, to avoid confusion with `hidden` systems.`
  And yes, while I *will* be temporarily changing them to `hidden` just to do `splitregion` runs, I'll change them back afterward. Cleaner commit history; don't make unnecessary changes till yadda yadda.
* I also like how the existing names all had "Italy split" in their full name. Makes it easy to search for the systems I want in my text editor. Didn't really work for me having a bunch of "split hidden system"s smushed together in one place in the file, with no good string to search for. Took the eyes a minute to focus in on'em. Hope that's OK.
* What I envision with this is that I'll be the only one to have to muck about with enabling/disabling these systems as we move forward.
  **More on that in the forum...**


michihdeu:
> • systems.csv entries. Why not just hold off on making unnecessary changes till it's time to make the necessary ones.

How can we test the systems with `tm/fast/datacheck` instead?

si404:
OK, looking, you are doing something entirely different to what I thought you were doing. I've understood this. I'll stay away rather than trying to help, given you get annoyed when I do - even when its on things you have said were a good idea like itae!

>This wasn't changed in https://github.com/TravelMapping/HighwayData/pull/7810, but should zafrtr have ZAF for its country instead of AFR?

Yes, for the same reason as I created itae you said "things could in fact be organized exactly as Si describes here." and thus I reorganised to be like that! zafrtr was already in that format.

yakra:
I think we're talking about different ideas without realizing we're talking about different ideas, and that's causing some miscommunication & confusion.

In that context, if I respond to a question or comment, I'm talking about what I'm talking about, and the reader might think I'm talking about the (different) thing they were talking about.
If I foresee that confusion and want to head it off in advance, I can qualify my comments with "*This* is what I'm talking about. Were you talking about this other thing instead? *If so...*"
Either way, communication gets messy... :(

---

To start out, I'll emphasize that the data in Italy right now is what we had in France right before the FRA split, long before `hidden` systems were thought of:
* Existing unsplit data in the existing systems that have always been there, whole country to a .wpt.
* Commented-out systems with the future *split* routes, ready to be spliced into the main systems once we're good to go.

This is what I'll be [running sanity checks](https://forum.travelmapping.net/index.php?topic=2976.msg13680#msg13680) on.

---

@michihdeu wrote:
> How can we test the systems with `tm/fast/datacheck` instead?

This question is complicated by the "Are we talking about the same thing?" factor.
* **For you** specifically, You won't need to do anything different, as you're not working on the ITA split.
* **In general?**
  * If preparing to split a system, just don't bother with the commented-out systems.
    Leave them commented out as you would for a production update. (And in doing so, they won't hurt anything.)
    I can do that instead, as part of running `--splitregion`, generating the [.log files](https://forum.travelmapping.net/index.php?topic=2976.msg13629#msg13629) and posting issues to the forum.
  * After a split has gone live, also do as you would normally.
    The temporary split systems will be gone & no longer an issue.
    Any `hidden` systems will be live, uncommented, and treated normally.
  * WRT my comment about systems.csv you quoted, the only changes that will need to happen are those for the changeover itself.

---

@si404, Please don't mistake my verbosity for annoyance (Okay, *this* time. I did [grumble](https://forum.travelmapping.net/index.php?topic=4781.msg35458#msg35458) that other time).
My intent was to explain what I was doing and why. And sometimes communication can be messy.

I never said itae was a good idea.
I assume you said that based on your quote from me, "[things could in fact be organized exactly as Si describes here.](https://forum.travelmapping.net/index.php?topic=4781.msg35458#msg35458)".
That was in reference to a way to potentially organize `hidden` systems in the future, once those become a thing.

But what we're dealing with now are still the split systems, the temporary stuff waiting in the rafters to take the place of of the `active` routes, just like we had for France.

When I think "itae", I think of itae as it was included in #7810.
This was exactly the same as eure2 was before creating it, minus the BIH routes and the E74 split, just with a different system name.
*(showing my work / saving for the future; you can ignore this bit)*
```sh
diff <(git show 8506395:./eure2.csv | grep -v BIH | while read line; do sys=$(echo "$line" | cut -f1 -d';'); echo "$line" | sed "s~^$sys;~xyz;~"; done) <(git show c778814:./itae.csv  | while read line; do sys=$(echo "$line" | cut -f1 -d';'); echo "$line" | sed "s~^$sys;~xyz;~"; done); diff <(git show 8506395:./eure2_con.csv | fgrep -v 'bih.' | while read line; do sys=$(echo "$line" | cut -f1 -d';'); echo "$line" | sed "s~^$sys;~xyz;~"; done) <(git show c778814:./itae_con.csv  | while read line; do sys=$(echo "$line" | cut -f1 -d';'); echo "$line" | sed "s~^$sys;~xyz;~"; done)
```
Thus this makes it one of the temporary, pre-split, "split" systems.
Which is actually what I *want* in order to do the sanity checks...
***Except***, the name change from `eure2` breaks the concurrency checker due to [the way it was implemented](https://forum.travelmapping.net/index.php?topic=2976.msg13691#msg13691).
That's why I reverted the system names.

---

What I meant my by zafrtr comment was, should the systems.csv line be changed from
`#zafrtr;AFR;SADC Regional Trunk Roads (split ZAF);TMlightsalmon;3;preview` to
`#zafrtr;ZAF;SADC Regional Trunk Roads (split ZAF);TMlightsalmon;3;preview` instead?
I didn't look at the contents of the system itself at all.

yakra:
Many thanks for the merge.
[On to the forum](https://forum.travelmapping.net/index.php?topic=6532.msg36136#msg36136), to start digging into the datachecks.

PR #7822, opened 2024-10-20
AB fixes from forum

yakra:
https://forum.travelmapping.net/index.php?topic=6536
https://forum.travelmapping.net/index.php?topic=6537


yakra:
oops hold up

PR #7881, opened 2024-11-02
ON: Correct transition between TCHGeo & TCHCen

rickmastfan67:
https://forum.travelmapping.net/index.php?topic=6550.0

As long as people have a entry for ON-12 along the area where the two routes have been swapped, you'll still have it correctly marked as traveled.


yakra:
6-field .list entries where both points were outside of the range where these changes occurred will also be OK.

PR #7950, opened 2024-11-29
Finland E16

si404:


michihdeu:
@si404 I think we should rename the wp label of the intersecting routes too, shouldn't we?

si404:
If it was Norway/Sweden/Denmark/Belgium, sure. But not Finland, just as we don't in France, etc.

PR #7976, opened 2024-12-09
W end NH119

yakra:
ping @ajfroggie

### DIFF
--- a/hwy_data/NH/usanh/nh.nh119.wpt
+++ b/hwy_data/NH/usanh/nh.nh119.wpt
@@ -1,5 +1,5 @@
-VT/NH http://www.openstreetmap.org/?lat=42.851598&lon=-72.556087
-+x11901 http://www.openstreetmap.org/?lat=42.848097&lon=-72.545857
+VT/NH http://www.openstreetmap.org/?lat=42.849128&lon=-72.554242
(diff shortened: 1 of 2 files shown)


jteresco:
Seems I should wait for the corresponding change on the VT side, assuming VT 119 is moving to connect to the new bridge also.

yakra:
It will be, yes, but I figure it's not like there'd be a DISCONNECTED_ROUTE to worry about.

ajfroggie:
I'm waiting until I can get down there (hopefully later this month) to
verify whether 119 ends at 142 or if it'll be signed up to 5 as it ended at
5 before.  Haven't seen photos one way or the other and the literature is
unclear.

yakra:
...Which is fine.

IMO this isn't a dealbreaker for changing the NH side now.
There's NH119 which does its thing in NH, and VT119 which does its thing in VT.
No need have them both the same level of up-to-date. It's not like this is a single US route crossing the border.

jteresco:
Hate to break the graph connection, but I won't be making another archive set for a few weeks probably, so no big deal.  I'll bring this is now.

yakra:
Ah yes, there is that.
Didn't think of it, somehow!

PR #7983, opened 2024-12-11
Fixing I-86/NY 17/NY 34/PA 199 NMPs

Markkos1992:
@yakra There may be other NY NMPs for you to check on, but this one is for the ones by the NY/PA line at NY 34/PA 199.


Markkos1992:
I was thinking that these NMPs should be corrected at least since some of them are technically mine.

PR #8004, opened 2024-12-20
ROU A1 & A7 Buzau N - Focsani N openings

rmihailucian:
@michihdeu please perform datacheck on this one. The last segment of A7 opens today, between Ramnicu Sarat and Focsani Sud.


michihdeu:
Data check successful.

PR #8030, opened 2024-12-27
Part 1 of 2024 GA updates

ntallyn:
Part 1 of 2024 GA updates


rickmastfan67:
@ntallyn, did you see my comments in the fourm about the I-75 Business
routes?  You might want to comment on it before this is pulled in.

Same goes for GA-201.

Thanks.

jteresco:
I'll hold off on this for now, then.

ntallyn:
Patch submitted. Re-added I-75 BL Valdosta; renamed US41_S point on GA201. Added minor realignment of US84/GA38 in Bainbridge (was in my queue before I saw these comments).

rickmastfan67:
Those GA fixes look good to me now.

However, I'm not sure what the heck is happening with the 'updates' csv
file. Several non-Georgia updates are marked for deletion, like 25+!

Not sure how we want to fix that issue.

ntallyn:
I merged into my branch before I started, but forgot to do the same before I pushed. I think it's fixed now.

rickmastfan67:
Just checked it again. Still 25+ entries marked for deletion. Mostly stuff
that was added in the last few days sadly.

ntallyn:
I think I have it fixed now.

rickmastfan67:
Looks good on my end now. :)

Jim, I think we're now safe to have this pulled in tonight when you're
ready to do the update.

rickmastfan67:
@ntallyn new issue with GA-201 discovered.  A point that should have been removed with the reroute wasn't, and thus causing a jump back to the 'old' alignment.

Just deleting the 'VarRd_S' point should fix this issue in a separate PR.  But it's something that should be addressed ASAP.

ntallyn:
Looks like I fixed it in the waypoint editor but failed to copy it back to the .wpt file. It's in a separate pull request.

Incidentally, I think the updates.csv issue was that I had the file open in the editor when I merged from upstream/master, so when it said that another file had modified the file, I ignored it. Oops on me.

PR #8069, opened 2025-01-10
GB changes (and NM)

si404:


michihdeu:
@si404 2025-10-01 -> 2025-01-10

PR #8080, opened 2025-01-13
email from Bickendan

si404:


michihdeu:
@si404 we don't make systemupdates entries for devel

michihdeu:
I'll remove them.

si404:
> I'll remove them.

No - I already did locally (and I pushed it to my branch so you could see I had). I'm going to get in a mess if you do!

PR #8082, opened 2025-01-13
srbrb review pt 1

si404:


michihdeu:
@si404 thanks. I think you missed this from the first post:
`R322: B124 -> R124`

PR #8087, opened 2025-01-15
POL + SRB (+IND+ENG)

si404:


michihdeu:
@si404 thanks for the update. You've extended R351 but did not truncate R350 (or you did not process R350):

```
R350:
Replace +X920202 by a visible wp at the next junction to the south
Truncate route to R351
```

PR #8147, opened 2025-02-11
Point Requests and US 85 Cleanup (off I-25)

the-spui-ninja:


hwy_data/CO/usaco/co.co066.wpt
```diff
@@ -1,6 +1,6 @@
 US36 http://www.openstreetmap.org/?lat=40.212252&lon=-105.249088
 75thSt http://www.openstreetmap.org/?lat=40.203327&lon=-105.178017
-95thSt http://www.openstreetmap.org/?lat=40.203325&lon=-105.131087
+95thSt/HovSt +95thSt http://www.openstreetmap.org/?lat=40.203325&lon=-105.131087
```
rickmastfan67, line 3: Pretty sure we don't use two different road names in labels.  We either choose 1 or the other.

the-spui-ninja: I've done that before where the roads are roughly of equal importance (Diamondville WY is the first place that comes to mind). I remember seeing something about it at one point but I can't find it in the manual now. I can change these when I notice them if you want.

rickmastfan67: I could have sworn we had a rule on this case as well.

Closest would be: https://travelmapping.net/devel/manual/wayptlabels.php#truncate

Maybe it's time we add an official rule on this, as I've been told to just go with 1 the few times I've done it in the past.

michihdeu: Ohne road name only, please!

the-spui-ninja: Sounds good, will change this one and the US 30/189 one in WY now (since I know where they are). I'll remove these (if there are more) when I find them looking around.

PR #8160, opened 2025-02-13
CT11 exit numbers

yakra:
No broken labels for this renumbering. Huzzah!
And now back to AARoads & ShadyJay's flickr, to see far along CT2 is.
I-384 has been on my radar, but I'm holding off till there's photo evidence of **all** exit numbers, or I can check it out in person.


rickmastfan67:
@yakra Did you completely forget about the ['fixes' mentioned here](https://forum.travelmapping.net/index.php?topic=4833.msg27029#msg27029) that also needed to be done to CT-2 that you promised back in 2022?  Would be nice to _FINALLY_ get that thread marked as solved.

Also, should *5B be converted to a mileage based *2C to prevent any 'possible' future ramps that might need that number?  '5B' isn't in use, so, no issues would be caused.

PR #8257, opened 2025-03-17
CA and NV additions and cleanup

cl994:


rickmastfan67:
I honestly think the new 'file name' for the NV airport connector is too long.

Thinking 'hraircon' would have been better.  Or even 'hracon', following the label naming standards due to something like this not being in the manual.

cl994:
> I honestly think the new 'file name' for the NV airport connector is too long due to having '4 words' in it.

Yeah, I agree. Renaming to "hraircon"

PR #8281, opened 2025-04-02
ITA+SCT

si404:


michihdeu:
@si404 you forgot truncating A13Fer

si404:
@michihdeu Fixed that locally this morning.

This is the problem, currently, with Italy - I have to change two files and I must have only changed the split one.

PR #8393, opened 2025-05-27
Update of India files

Bickendan:
Upload and update of India routes


si404:
There's no .csv changes so it depends on the deletions (ie if they were non-existant routes) if there's any issues with update.

I'll pull it, and run datacheck.

si404:
Datacheck successful.

However, none of the added routes will appear in the browser as they aren't listed in the csvs in the _systems directory

PR #8403, opened 2025-05-29
LA Route Truncations

cenlaroads:
LA 93-1, 342, and 3246 truncated
LA DOTD Road Transfer maps and GSV imagery confirm truncations


cenlaroads:
Datacheck successful

PR #8405, opened 2025-05-30
Add files via upload

Bickendan:
OR 11 and 339 changes


Bickendan:
The new ones are correct.

si404:
I've fixed the merge conflicts though I don't know how they got there - perhaps my removal of unused alt labels? Though that helped because the addition of those alt labels meant I could find the bits to get rid of easier

PR #8421, opened 2025-06-06
WA Updates

compdude:
-Update WA 99, WA 526, and WA 527 at  interchange with I-5 to reflect that each of the highways end at the intersection with each other west of I-5 instead of ending at I-5 
-Add HouWay point to WA 900, recentering several other waypoints 
-Recenter points for I-5, I-405, and I-90 at interchanges with WA 527 and WA 900 
-Extend WA 509 from I-705 to Pacific Avenue, add updates entry for this change 
-Recenter points on I-705 and WA 518 where they intersect with WA 509

Resolves these forum threads: https://forum.travelmapping.net/index.php?topic=4450.msg24866#msg24866 and https://forum.travelmapping.net/index.php?topic=5436.0


compdude:
Okay, now that I've added an updates entry for all my changes, this should be good to go.

PR #8453, opened 2025-06-29
Japan national highways 40-49

denishanson:
Second pass thru national highways 40-49 and their concurrencies.
Added wps for GSJs
Converted some shaping wps to named wps


denishanson:
Passed datacheck

PR #8465, opened 2025-07-03
ROU A0 and A1 openings

rmihailucian:


rmihailucian:
@si404 @michihdeu all will be added during the next days, I know about them, just that it's harder to find time for TM in the last years.

PR #8492, opened 2025-07-15
ny.ny266: fix error log entry

rickmastfan67:
Just wanted to fix the error log entry of this, especially since @yakra sorta gave a green light to it in https://github.com/TravelMapping/DataProcessing/issues/674#issuecomment-3032535580.


rickmastfan67:
Thread in question: https://forum.travelmapping.net/index.php?topic=6376.0

PR #8512, opened 2025-07-22
forgot to update the update date

the-spui-ninja:


yakra:
Should be 2025, right?

the-spui-ninja:
I wish there was a facepalm emoji

PR #8515, opened 2025-07-22
Update Japan N5-N20

denishanson:
Added wps for GSJs
Added wps
Converted shaping wps to named wps
Simplified wp names


denishanson:
Passed datacheck

PR #8535, opened 2025-07-28
Mexico

neroute2:
data check successful


neroute2:
> There are SEVERAL broken links on the 'updates' page.
> 
> 2025-07-27 | (Mexico) Morelos | MEX91D | mexmor.mex91d | new route (was MEX95DIgu)
> 
> This is just one example.

Oops, forgot a leading zero.

PR #8579, opened 2025-08-26
Supermassive datacheck

si404:
This should create a decent graph set with lots of errors removed (only thing not done was mark NMP FPs, but all true positives in my regions fixed).


hwy_data/IL/usaus/il.us136.wpt
```diff
@@ -63,5 +63,6 @@ IL1/119 http://www.openstreetmap.org/?lat=40.299494&lon=-87.658110
 PolRd http://www.openstreetmap.org/?lat=40.182349&lon=-87.629828
 VooSt http://www.openstreetmap.org/?lat=40.145814&lon=-87.630215
 US150/1 +US150/IL1 http://www.openstreetmap.org/?lat=40.124389&lon=-87.635665
+VerSt http://www.openstreetmap.org/?lat=40.124439&lon=-87.630000
```
rickmastfan67, line 66: This label needs to be something different. Adding it caused a 'Duplicate Label' error in US-136's file.

rickmastfan67: @si404 I guess you missed this because I forgot to tag you in it when you fixed the IL US-30 one above.

Thankfully, it's not in-use, but should be addressed before that can happen.

`il.us136 | VerSt | DUPLICATE_LABEL |   | This is always a true error and cannot be marked false positive.`

PR #8622, opened 2025-09-23
fixes

si404:


si404:
Fixed that this morning, but forgot to commit (or at least sync the commit)
and am now on the road for a week.

I'll see if I can fix, but if you have the opportunity please do.

michihdeu:
no problem, it can wait a week.

michihdeu:
> @si404
> 
>     * there is a concurrency issue with A68 and N232Gal since you forgot truncating N232Gal wpt file (only update entry)
> 
>     * A68Car is fukced up, CamEnm should be the last wp, not the first one.

@si404 you missed that with  #8637

PR #8642, opened 2025-10-05
Create qc.qc138rom

ovoss:
New QC 138 segment, part of in-progress project to extend QC 138 from Kegaska east to La Romaine, and ultimately to Blanc Sablon.

Updates entry later, in separate pull request.


ovoss:
Jim, something went wrong with this PR. New route apparently didn't make it into the HB.

PR #8647, opened 2025-10-06
RI route 1A Westerly

IMGoph:
Added points to connect to locally important streets around Watch Hill and Misquamicut.


IMGoph:
I am not sure what that second commit request is. I thought I did things the same way I have for others over the past week, but something seems odd. Let me know if this needs to be redone.

PR #9054, opened 2026-03-23
MC+SVK+POL

si404:


michihdeu:
@si404 https://travelmapping.net/devel/datacheck.php?rg=svk

Either add `Komarno` to both csv files, or rename the route so that we get rid off `abbrev` field entry. I know that the latter would trigger 3 users with deprecated route errors.

si404:
Or just leave it, because it's not a real error and it is hurting nobody?

michihdeu:
C'mon....

si404:
Don't be a dweeb. I'm not breaking people's list files to remove a datacheck 'error'.

PR #9120, opened 2026-04-24
ENG + FL

si404:


rickmastfan67:
@si404 Please at least tag me in the pull requests if you change something on an active FL highway.  I keep offline files that I use as the primary backups for the state.

I have no problems putting your changes into them, just need to be alerted to it when it happens to keep them synced in case I need to change something else in the file (like if a new signed highway is opened).

I only noticed this change to US-1 due to getting ready to push an update to a separate US Highway.

PR #9159, opened 2026-05-24
Some NY updates from the forum

jteresco:


hwy_data/NY/usany/ny.ny022.wpt
jteresco, line 1: Of interest to @alexgotz

jteresco:
That's it for now.  I think I've gotten everything in NY reported at least 10 months ago either taken care of here, or bumped the threads for things I was less certain about.  It's not how I planned to spend a few hours on a rainy holiday weekend afternoon, but I enjoyed playing highway data manager for the first time in a long time.  Hopefully I didn't break anything and I don't think I did anything @yakra would object to too much in his region.

jteresco:
Two datacheck fixes were needed after this, both fixed in the repository and NY should be off the datacheck errors list again tomorrow night.

PR #9171, opened 2026-05-26
Delete LA 502, LA 3169 (West Colfax), and LA 3169 (East Colfax)

cenlaroads:
https://forum.travelmapping.net/index.php?topic=7564.0

LA 3169 confirmed deleted by LA DOTD email and GSV


cenlaroads:
Datacheck successful

PR #9173, opened 2026-05-27
ENG + tourist routes

si404:


michihdeu:
@si404 2 out of 5 usatr changes to RI, 1 out of 16 for OK.... I'm happy that you work on it but I don't get why you only pick out 3 (plus VA that was not listed)... I was about to split the thread but it's difficult when some changes are already made but not all of one region.

https://forum.travelmapping.net/index.php?topic=7039

PR #9195, opened 2026-05-31
Add OR 339 Updates Entry

Markkos1992:
https://forum.travelmapping.net/index.php?topic=6950.msg38438#msg38438


jteresco:
<pre>
[5.9] ABORTING due to 1 errors:
1: Could not parse updates.csv line: [2025-05-30;(USA) Oregon;OR 339;or.or339;Truncated from OR 11 through Milton-Freewater to the northern city boundary at LambSt; route reversed to match Oregon north-south Highway mileage convention.], expected 5 fields, found 6
</pre>

Don't forget to run datacheck even for simple updates entries.  The text description can't have a semicolon since it is interpreted as another field delimiter.

I'll fix it and re-run.

PR #9201, opened 2026-05-31
Caucasian redo

si404:


michihdeu:
@si404 the ARM E117 entry seems to be a copy-paste error.

michihdeu:
and several disconnected route errors: https://travelmapping.net/devel/datacheck.php
but thanks for the update 😄

si404:
> and several disconnected route errors: https://travelmapping.net/devel/datacheck.php but thanks for the update 😄

I can check datacheck all by myself, thanks.

> @si404 the ARM E117 entry seems to be a copy-paste error.

???

michihdeu:
The update link to ARM E117 dies not work, the update Description is the same as for the line before.

si404:
I still don't know what the actual problem is here. Having no problems with the link and the description of the change is spot on.

michihdeu:
Sorry, AZE, not ARM

si404:
That makes sense. Was a typo - E119 not E117.

PR #9212, opened 2026-06-01
TX:  Extend Westpark Tollway; OK:  Correct waypoint on OK 88, Mark FPs

cenlaroads:
https://forum.travelmapping.net/index.php?topic=6837.msg1#new

https://forum.travelmapping.net/index.php?topic=7434.0


cenlaroads:
I think these are changes that yakra would find uncontroversial

Correct waypoint, mark sharp angle FP, extend Westpark Tollway based on multiple sources

PR #9239, opened 2026-06-09
MB39 Realignment/NY17K Point Request

cockroachking:
https://forum.travelmapping.net/index.php?topic=7614.0
https://forum.travelmapping.net/index.php?topic=7611.0


cockroachking:
Ping:
@ovoss (MB39)
@mr-matt-k (NY17K)

ovoss:
On MB 39, sure that PR 596 (which we call MB 596) didn't absorb part of old MB 39? The MB 39 realignment de-clinches that route for me, but it looks from OSM that PR 596 now includes part of the old MB 39 alignment, which I've traveled so I can at least salvage that km-age.

That would mean MB 596's north endpoint would need to be moved to the new intersection, with old MB 39 point in the MB 596 file renamed OldMB39.

jteresco:
Best I can tell, the answer to @ovoss's question doesn't change what was done here so I am going ahead.

cockroachking:
@ovoss See https://forum.travelmapping.net/index.php?topic=2344.msg42007#msg42007 for my explanation. In short, I'm not 100% sure and canmbp is not my system* to make judgement calls on.

*Unless we are now allowing 3rd parties to help out with stagnant preview systems. I was under the impression that the recent push was for active systems only.

PR #9276, opened 2026-06-22
Split US 460 / create US 460 Alt

mapcat:
Kentucky route moved onto KY 3174 and split from VA/WV route at state line
US 460 Alt (Shelbiana) covers former routing in Kentucky
Waypoint requests and minor changes in MI, OH, PR, TN


mapcat (commented):
NMP fixes (KY, MI, TN)

PR #9383, opened 2026-08-03
cannl Peer Review Restart

cockroachking:


hwy_data/NL/cannl/nl.nl450.wpt
```diff
@@ -2,7 +2,7 @@ NL1 http://www.openstreetmap.org/?lat=48.910173&lon=-57.895575
 CplPinDr http://www.openstreetmap.org/?lat=48.928139&lon=-57.927132
 SunDr http://www.openstreetmap.org/?lat=48.930592&lon=-57.988758
 OConDr http://www.openstreetmap.org/?lat=48.946773&lon=-57.990861
-LewPkwy_N http://www.openstreetmap.org/?lat=48.948717&lon=-57.990088
+NL450A +LewPkwy_N http://www.openstreetmap.org/?lat=48.948645&lon=-57.990038
```
rickmastfan67, line 5: Since this system is in preview still, we don't have to keep old labels, even if in use.  People expect labels could change till it's activated.

jteresco: Since they do no harm, and this system was in preview for a very long time, I'd say include the old labels if they help any users.

michihdeu: @rickmastfan67 there is no downside in keeping alt labels for preview systems whereever possible. si404 and me do it for years....

cockroachking: I'll also add that The Manual implies we should as well: https://travelmapping.net/devel/manual/sysnew.php#reviewbreak

PR #9384, opened 2026-08-03
Turnpikes

si404:
This will need datacheck as I can't do it thousands of miles from home...

PR #9386, opened 2026-08-03
fixes for #9384

si404:
Hopefully this should do it. I've added x banners to the relevant turnpikes that can be removed upon activation (we'd also need to work out what, if any, altlabels need adding).

Obviously needs checking...


michihdeu:
1: Could not find Route matching ConnectedRoute root ok.indnattpkx in system usaoktpk.
2: No valid roots in usaoktpk_con.csv line: usaoktpk;IndNatTpk;x;Indian Nation Turnpike;ok.indnattpkx
3: [Errno 2] No such file or directory: '/fast/home/michih/HighwayData/data/OK/usaoktpk/ok.indnattpk.wpt'
4: usaoktpk.csv: root ok.indnattpk not matched by any connected route root.

PR #9411, opened 2026-08-15
LA:  Delete LA 182-2, Truncate LA 339

cenlaroads:
https://forum.travelmapping.net/index.php?topic=7720.0


cenlaroads:
Datacheck successful

PR #9418, opened 2026-08-17
Updated files for OR by email from @Bickendan

jteresco:


jteresco:
I'd like @Bickendan and one other person to check this out before I merge in to make sure the changes as I put them in look accurate.

michihdeu:
Changes to `usanht`... I think @si404 should look into it...

jteresco:
Yes, I'll wait until @si404 has a chance to take a look before merging.

si404:
I have no problem - seems to be rebuilding the routes in the style used by Bickendan so as to make it easier to deal with concurrencies.

PR #9421, opened 2026-08-18
Fix broken link

rickmastfan67:
(added missing '0')

Fixed issue in #9417.

@ntallyn


ntallyn:
Thanks for the catch.

PR #9428, opened 2026-08-20
AR/KS/MA/NB/NE/TX Unused Alt Labels

cockroachking:


cockroachking:
Well that's interesting. Concurrency check failed even though I didn't touch any coordinates, only labels...

jteresco:
It actually was successful but for some reason it flagged the success as a failure.  I'm still considering this a win because it gave the correct output (that no .wpt coordinates were modified).  A very fixable glitch I'm sure.

PR #9434, opened 2026-08-21
BY/TH

michihdeu:


michihdeu:
@jteresco the failed coordinate consistency check is a FP. I've moved the route onto a new alignment and thus deleted a point that still exists for the other route. The new wp looks like a moved wp for your check but it isn't.
I think it doesn't matter that the check failed. I could easily check that everything is intended.

jteresco:
Yes, that's good.  Exactly a situation that the author of a change knows is intentional and can be ignored, as you did.

PR #9448, opened 2026-08-24
BB Cottbus

michihdeu:
fix for https://github.com/TravelMapping/HighwayData/pull/9444 I obviously forgot copying the updated route to the wpt file


michihdeu:
Consi check FP

PR #9465, opened 2026-08-31
Unhiding Exits 41, 55, and 58

Bickendan:
Trailhead access for hikers and bikers


Bickendan:
Changed OR 281's point to match US 30 (and OR 35)

jteresco:
Assuming the coordinate check was a false positive but please check to be sure @Bickendan

PR #9466, opened 2026-08-31
Unhiding Exits 41, 55, and 58

Bickendan:
Trailhead access for hikers and bikers


jteresco:
Should I merge both #9465 and this one?

Bickendan:
Yes. The US 30 and OR 281 fix both had an error I couldn't figure out, but
both I-84 and US 30 needed the points unhidden.

PR #9467, opened 2026-08-31
Resyncing points with US 30 and OR 35

Bickendan:


jteresco:
@Bickendan Never mind on my comment about checking the other, I see you fixed it here.  It would be better to include this new commit on the previous pull request which would re-run the consistency checker so you could see the check failure change to a pass.  (Plus keeping things in a single pull request saves me a few clicks and checks when I am merging things in before the site update.)

PR #9478, opened 2026-09-03
UT 148 truncation / move rest to usanp

Duke87ofST:


jteresco:
coordinate check false positive?

Duke87ofST:
It's a false positive yes - is there a standard method of marking it as one?

jteresco:
@Duke87ofST No need to mark, as the change will never be examined again once this gets merged in.  It's a check on a difference, not on the new state of the files.

michihdeu:
@Duke87ofST I think we should simply notify in the conversation that it's a FP so that we know that the PR can be merged. Thanks.
Si simply notifies: `CCC FPs`

PR #9492, opened 2026-09-06
HUGE Puerto Rico update 

TBKS1-Road:
This took a lot longer than I thought it would. Multiple failed datachecks which have since been fixed, and the amount of routes added, changed, and removed took quite a while to get everything included. No promises that this is perfect but this finally passed datacheck later.

https://forum.travelmapping.net/index.php?topic=5615.0

I will be out of town for about a week and a half starting tomorrow. If there's any issues, please let me know!


TBKS1-Road:
Unfortunately, it appears that this did not check properly, so I'm going to close out this pull request. The thread will still be open but I don't think I'll be able to get to this today (and quite frankly this was an absolute nightmare on my part, someone with more developer experience might be better off adding everything in.)

I'll be on the road for about a week and a half starting tomorrow.

PR #9499, opened 2026-09-07
West Bengal State Highway revisions

Bickendan:
Initial pass revision

WB 12 will need to be split into three files later (Siliguri, Nepal, and Darjeeling)


jteresco:
waiting to make sure the CCC report is all FPs (I suspect they might not be).

si404:
I've removed the files asked for, and re-ran the CCC. It's still giving the obsolete files, so it will need re-running after the next site update to remove those.

There are some other things that don't involve the obsolete routes (now removed). I don't know whether or not they are FPs.

michihdeu:
> There are some other things that don't involve the obsolete routes (now removed). I don't know whether or not they are FPs.

@Bickendan

Bickendan:
Mark the remaining coordinate discrepencies as FPs, as they should coincide with the NH border points with surrounding states.

PR #9504, opened 2026-09-09
PR Thread Part 1

cockroachking:
https://forum.travelmapping.net/index.php?topic=5615.0


cockroachking:
Coord check errors are FPs.

PR #9509, opened 2026-09-11
IRN/PR/PRT/WA Updates

cockroachking:
https://forum.travelmapping.net/index.php?topic=7819.0


hwy_data/IRN/irnf/irn.f021.wpt
```diff
@@ -1,3 +1,3 @@
-F21 http://www.openstreetmap.org/?lat=35.607112&lon=51.508013
+R11/44 http://www.openstreetmap.org/?lat=35.607112&lon=51.508013
```
michihdeu: @cockroachking I think it should be R11/R44.

https://travelmapping.net/devel/manual/wayptlabels.php#slash

PR #9513, opened 2026-09-12
CA I-380 and I-8 BL El Centro extended east

cl994:
- CA I-380 extended east to N Access Rd to match Caltrans logs
- CA I-8 BL El Centro extended east to match signs

### DIFF
--- a/updates.csv
+++ b/updates.csv
@@ -8709,6 +8709,7 @@ date;region;route;root;description
 2015-09-03;(USA) Arkansas;US 62 Business (Prairie Grove);ar.us062buspra;Added route
 2015-09-03;(USA) Arkansas;US 62 Business (Rogers);;Deleted route
 2015-09-03;(USA) Arkansas;US 63 Business (Jonesboro);;Deleted route
+2026-08-13;(USA) California;I-8BL (El Centro);ca.i008blelc;East end moved to I-8 Exit 23 to match signs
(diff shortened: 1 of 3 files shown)


jteresco:
Looks like the CCC might be a valid problem.  If not, say so here and I'll merge.   @cl994

cl994:
CCC is a false positive, it's mad that I changed the intersection of I-8 and the BL to another interchange. Should be good to merge now with added updates as on forum.

rickmastfan67:
@cl994 missing an update entry for the I-380 change since it was extended beyond Airport Blvd.

updates.csv
```diff
@@ -8709,6 +8709,7 @@ date;region;route;root;description
 2015-09-03;(USA) Arkansas;US 62 Business (Prairie Grove);ar.us062buspra;Added route
 2015-09-03;(USA) Arkansas;US 62 Business (Rogers);;Deleted route
 2015-09-03;(USA) Arkansas;US 63 Business (Jonesboro);;Deleted route
+2026-08-13;(USA) California;I-8BL (El Centro);ca.i008blelc;East end moved to I-8 Exit 23 to match signs
```
rickmastfan67, line 8712: Just noticed, the date on this needs to be fixed too.  Says 08/13 when it should be 09/13. ;)

PR #9516, opened 2026-09-13
I-380 CA

cl994:


rickmastfan67:
You had added I-380 in as I-308, thus the link for it was broken on the updates page.  I just submitted a fix for that. ;)

PR #9520, opened 2026-09-13
OR 282

Bickendan:
Points added.


Bickendan:
FP from shapepoint proximity.

PR #9523, opened 2026-09-15
ENG + NSW + SVK + OR

si404:


si404:
CCC (has that been improved? - I expected more errors to be flagged up) is an FP

michihdeu:
> CCC (has that been improved? - I expected more errors to be flagged up) is an FP

The check refers to [coord-check.yml](https://github.com/TravelMapping/HighwayData/blob/master/.github/workflows/coord-check.yml) (last change 2026-08-20) and calls [ccc.py](https://github.com/TravelMapping/DataProcessing/blob/master/CoordConsistencyCheck/ccc.py) (last change: 2026-09-04)

PR #9527, opened 2026-09-18
ENG + turnpikes

si404:


hwy_data/PA/usapatp/pa.patp043fai.wpt
```diff
@@ -0,0 +1,8 @@
+WV/PA http://www.openstreetmap.org/?lat=39.720738&lon=-79.822090
+2 http://www.openstreetmap.org/?lat=39.741209&lon=-79.810873
+4 http://www.openstreetmap.org/?lat=39.767439&lon=-79.792293
++X1 http://www.openstreetmap.org/?lat=39.781959&lon=-79.774475
++X2 http://www.openstreetmap.org/?lat=39.802629&lon=-79.783308
+8 http://www.openstreetmap.org/?lat=39.818636&lon=-79.776438
```
rickmastfan67, line 6: @si404 As I mentioned in the Turnpike thread, Exit 8 up to US-119 is NOT part of the PA Turnpike.  That segment is 100% PennDOT owned/maintained and is signed as such with the normal PA shield (they built it as a demonstration project since the Turnpike didn't have the money at the time).  The Turnpike has 100% nothing to do with it (hence why the US-119 exit doesn't have a number).

PR #9528, opened 2026-09-19
Create indsk

Bickendan:
directory for Sikkim state highways


jteresco:
Looks like a blank file?

Bickendan:
Meant to be the directory for all the IND-SK state highways.

jteresco:
A common point of confusion with git is about creating directories.  Empty directories don't exist in git, they just get created when you commit files or subdirectories to them and go away if all the files and subdirectories within get removed.

PR #9532, opened 2026-09-23
Japan updates

denishanson:
E6 new IC 3-1
Part of N203 changes


denishanson:
Coordinate consistency check doesn't seem to be able to handle the N203 changes (delete N203Kyu, move N203 onto N203Kyu, start N203Tak at former end of N203Kyu, movement of named WPs).  I will try doing it in pieces.

michihdeu:
CCC does not have to complete successfully. It's an imperfect check. You can simply notify `CCC FP` in the conversation when you are sure it's FP.

jteresco:
^^ exactly- this is just a tool to help find problems that creep into the data when a route is modified without a corresponding modification to another route that uses the exact same lat/lng point.  Just look at the CCC output and if what it reports doesn't look like it's concerning to you, you can just put a comment in the PR that things are fine and I or someone else will merge in like always.

